PR #5, opened 2021-03-15
Handle signatures

germartinez:


src/EthersSafe.ts
```diff
@@ -0,0 +1,111 @@
...
+      safeTransactionData.refundReceiver,
+      safeTransactionData.nonce
+    )
+    return txHash
+  }
+
+  async confirmTransaction(safeTransaction: SafeTransaction): Promise<void> {
+    const owners = await this.getOwners()
+    if (
+      owners.filter((owner: string) => owner.toLowerCase() === this.#signer.address.toLowerCase())
```
mikheevm, line 58: [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) seems to be a better fit there

also maybe it's worth to extract function for comparing strings as an utility function, most likely we'll have to use a lot throught the project

src/utils/transactions.ts
```diff
@@ -0,0 +1,77 @@
...
+    signers.forEach((signerAddress) => {
+      const signer = this.signatures.get(signerAddress)!!
+      staticParts += signer.staticPart(/*baseOffset + dynamicParts.length / 2*/)
+      dynamicParts += signer.dynamicPart()
+    })
+    return '0x' + staticParts + dynamicParts
+  }
+}
+
+export function makeSafeTransaction({
```
mikheevm, line 52: why do we need a separate function? why one can't just use `new SafeTransaction`

PR #6, opened 2021-03-15
Add read methods

germartinez:
Add read methods to access the data from the Safe contract

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,40 +29,33 @@ This is how executing a Safe transaction with off-chain signatures looks like.
 1. Create a Safe account with two owners and threshold equal two (this configuration is just an example):
 ```js
 import { ethers } from 'ethers')
-import EthersSafe, { makeSafeTransaction } from 'safe-core-sdk'
+import EthersSafe, { SafeTransaction } from 'safe-core-sdk'
 
 const provider = ethers.getDefaultProvider('homestead')
 const wallet1 = ethers.Wallet.createRandom().connect(provider)
 const wallet2 = ethers.Wallet.createRandom().connect(provider)
 
-const safe = await safeContract.setup(
-  [wallet1, wallet2], //owners
-  2, // threshold
-  AddressZero,
-  '0x',
-  AddressZero,
-  AddressZero,
-  0,
-  AddressZero
-)
+// Existing Safe address (e.g. Safe created via https://app.gnosis-safe.io
+const safeAddress = "0x<safe_address>"
+const safeNonce = <safe_nonce>
 ```
 
 2. Create a Safe transaction:
 ```js
-const tx = makeSafeTransaction({
-  to: safe.address,
+const tx = new SafeTransaction({
+  to: safeAddress,
   value: '0',
   data: '0x',
-  nonce: (await safe.nonce()).toString()
+  nonce: safeNonce
 })
 ```
 
 3. Generate the signatures with both owners:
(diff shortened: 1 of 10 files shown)


src/EthersSafe.ts
```diff
@@ -10,31 +10,89 @@ class EthersSafe implements Safe {
   #ethers: any
   #signer: Wallet
 
+  /**
+   * Creates an instance of the Safe Core SDK.
+   *
+   * @param ethers - Ethers v5 library
+   * @param signer - Ethers signer
+   * @param safeAddress - The address of the Safe account to use
+   * @returns The Safe Core SDK instance
```
mikheevm, line 19: shouldn't return also include a type? https://jsdoc.app/tags-returns.html

btw I think typescript supports jsdoc and can generate types from it, so maybe we don't need to duplicate them: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

README.md
```diff
@@ -1 +1,81 @@
...
+1. Create a Safe account with two owners and threshold equal two (this configuration is just an example):
+```js
+import { ethers } from 'ethers')
+import EthersSafe, { makeSafeTransaction } from 'safe-core-sdk'
+
+const provider = ethers.getDefaultProvider('homestead')
+const wallet1 = ethers.Wallet.createRandom().connect(provider)
+const wallet2 = ethers.Wallet.createRandom().connect(provider)
+
+const safe = await safeContract.setup(
```
rmeissner, line 38: Where is the `safeContract` coming from?

src/EthersSafe.ts
```diff
@@ -52,6 +110,24 @@ class EthersSafe implements Safe {
     return txHash
   }
 
+  /**
+   * Signs data using the current owner account.
+   *
+   * @param hash - The data to sign
+   * @returns The Safe signature
+   */
+  async signMessage(hash: string): Promise<SafeSignature> {
```
rmeissner, line 119: I am not sure about the name as the Safe also has a method called `signMessage`

src/utils/transactions.ts
```diff
@@ -49,6 +49,12 @@ export class SafeTransaction {
   }
 }
 
+/**
```
rmeissner, line 52: the baseOfsset in line 40 should probably be `signers.length * 65` as we need the offset in bytes not in chars.

rmeissner (approved):

README.md
germartinez: How would you represent an existing Safe account here? It doesn't make sense to add all the logic to explain where this is coming from.

germartinez: It is also weird that a direct interaction with the contract is needed before using the SDK. I guess at some point the SDK should allow to create Safes, right? Should this be prioritized?

src/EthersSafe.ts
germartinez: Is `signTransactionHash` better?

README.md
rmeissner: maybe just add a comment like
```js
// Existing Safe address (e.g. create from a Safe created via https://app.gnosis-safe.io
const safeAddress = "0x<some_safe_address>"
```

src/EthersSafe.ts
rmeissner: yes, this would make it a lot clearer for me :)

PR #17, opened 2021-03-29
Gas estimation and improvements

germartinez:
- Instantiate the SDK with a static method
- Standardize Safe transactions
- Estimate `safeTxGas`
- Calculate `gasLimit`
- Update tests


src/utils/index.ts
```diff
@@ -1,3 +1,6 @@
-export function areAddressesEqual(address1: string, address2: string): boolean {
-  return address1.toLowerCase() === address2.toLowerCase()
+export const sameString = (str1?: string, str2?: string): boolean => {
+  if (!str1 || !str2) {
```
rmeissner, line 2: but if both are `undefined` then they are the same :D

src/utils/gas.ts
```diff
@@ -0,0 +1,176 @@
+import { Provider } from '@ethersproject/providers'
+import { BigNumber } from 'ethers'
+import { GnosisSafe } from '../../typechain'
+import { sameString } from './'
+import { EMPTY_DATA } from './constants'
+
+// Parses the result from the error message (GETH, OpenEthereum/Parity and Nethermind) and returns the data value
+const getDataFromNodeErrorMessage = (errorMessage: string): string | undefined => {
```
rmeissner, line 8: Does this work with hardhat (or ganache)?

germartinez: No, it doesn't. That's the issue I'm having now

src/utils/index.ts
mikheevm: but why would it accept undefined as an argument? that doesn't make sense. the parameters have to be a string in order for the function to work properly, if one of the arguments you want to pass can be undefined, use an if statement and don't call the function

rmeissner: even if you don't accept `undefined`, same applies to the empty string ... if both are empty this should return `true` but as `!""` evaluates to `true` this function will return `false` if both strings are empty ... which is wrong according to the name.

mikheevm: If you don't accept undefined you don't need this check

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2020", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "target": "ES6", /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
mikheevm, line 3: I'd use ES2017 or newer, if a project needs to transpile it they'll do it. You can go es2020 to es5 with transpiling tools, but not vice versa. New ECMAScript standard bundles may have a smaller bundle size.

README.md
```diff
@@ -30,95 +30,94 @@ A Safe account with three owners and threshold equal three will be used as the s
...
 
 ### 1. Create a Safe transaction
 
 ```js
-const tx = new SafeTransaction({
+const tx = await safeSdk.createTransaction({
   to: safeAddress,
   value: '0',
   data: '0x',
-  nonce: safeNonce
```
mikheevm, line 57: Did you mean this PR when we talked about allowing to pass a custom nonce? https://github.com/gnosis/safe-core-sdk/pull/19#discussion_r619105464

if yes, how was it implemented?

src/EthersSafe.ts
```diff
@@ -1,22 +1,23 @@
...
 
 class EthersSafe implements Safe {
-  #contract: GnosisSafe
+  #contract: any
   #ethers: any
-  #ownerManager: OwnerManager
-  #moduleManager: ModuleManager
-  #provider: Provider
-  #signer?: Wallet
+  #ownerManager!: OwnerManager
```
mikheevm, line 17: what is the exclamation mark for and why did it become required?

README.md
mikheevm: Oh, I see `SafeTransactionDataPartial`. I could find in docs that you can pass custom nonce, though. Do you plan to add it?

src/EthersSafe.ts
germartinez: I'm using the property `strictPropertyInitialization` in the tsconfig.json file, which enables strict checking of property initialization in classes.
Because there are some async tasks to check in the constructor of the class, I'm using a static and async method `create` to return an instance of the class. The property `strictPropertyInitialization` returns the error `Property '<name>' has no initializer and is not definitely assigned in the constructor.` because the constructor is not used to initialize them and the assertion operator (!) can be used to ingore that (ans still have the property `strictPropertyInitialization` everywhere else).
Here you have more info: https://www.typescriptlang.org/docs/handbook/2/classes.html#--strictpropertyinitialization

mikheevm: Got it. I'd prefer usage of @ts-expect-error, though

mikheevm (approved):
buen trabajo

Issue #9, opened 2021-03-23
Handle write contract functions


germartinez:
- Add/remove/replace owners
- Change threshold
- Enable/disable Safe modules

PR #19, opened 2021-04-12
Handle write contract functions

germartinez:
Check #9

- Add/remove Safe modules
- Change threshold
- Add/remove/replace Safe owners
- Prevent contract errors with input validation in the SDK
- Tests


README.md
```diff
@@ -268,6 +276,88 @@ const tx = new SafeTransaction({
...
+})
+```
+
+#### Transaction to disable a Safe module.
+
+```js
+const tx = await safeSdk.buildContractCall({
+  method: SafeSettings.DISABLE_MODULE,
+  moduleAddress
+})
```
mikheevm, line 303: I'm curious why you chose the approach of having one generic method to perform safe contract calls. It feels too "raw" or low-level 🤔 I think it's more handy to have a method responsible for the call (enableModule, disableModule, etc). This buildContractCall could be used in the internal realization. You can find an example here: https://github.com/gnosis/safe-apps-sdk/blob/master/packages/safe-apps-sdk/src/eth/index.ts#L84 Probably, not the best one 🙈 I also found that in web3.js repo, they also use that approach to build methods for rpc calls

germartinez: This is a good comment :)

I had it implemented that way before refactoring to just just have one generic method. You can check here the commit and the methods I had previously:
https://github.com/gnosis/safe-core-sdk/pull/19/commits/35c0aa17107f10401434ca0230861048b673a7f1#diff-9d1172cb852904ea345f63e4470d48e4b305c14b39b6ef97202c28c356e58676L25

So it is this:

```js
getEnableModuleTx(moduleAddress: string): Promise<SafeTransaction>
getDisableModuleTx(moduleAddress: string): Promise<SafeTransaction>
getAddOwnerTx(ownerAddress: string, threshold?: number): Promise<SafeTransaction>
getRemoveOwnerTx(ownerAddress: string, threshold?: number): Promise<SafeTransaction>
getSwapOwnerTx(oldOwnerAddress: string, newOwnerAddress: string): Promise<SafeTransaction>
getChangeThresholdTx(threshold: number): Promise<SafeTransaction>
```

vs. this:

```js
buildContractCall(params: ContractCallParams): Promise<SafeTransaction>
```

I don't really have a strong opinion on this. I unified them because I thought the logic/concept (not the implementation) was almost the same: to get the encoding of a contract call data. So why not using just one function passing the contract method name as an argument?

Does this look too low-level? It may be.
What do you think @rmeissner? Do you also agree that is easier for the devs to have one independent method for each contract call?

rmeissner: Hmm I also don't have a strong opinion. Ethers allows you to do both approaches, right? Not sure what is best. Probably the one where autocomplete works the best :P

src/managers/types.ts
```diff
@@ -0,0 +1,49 @@
+export enum SafeSettings {
```
mikheevm, line 1: I can see an import in index.ts for this file:
```
import { ContractCallParams, SafeSettings } from './managers/types'
```

I couldn't find any other ones, so seems not all the types are used.

src/EthersSafe.ts
```diff
@@ -237,6 +253,106 @@ class EthersSafe implements Safe {
...
+   *
+   * @param moduleAddress - The desired module address
+   * @returns The Safe transaction ready to be signed
+   */
+  async getEnableModuleTx(moduleAddress: string): Promise<SafeTransaction> {
+    const tx = new SafeTransaction({
+      to: this.getAddress(),
+      value: '0',
+      data: await this.#moduleManager.encodeEnableModuleData(moduleAddress),
+      nonce: (await this.#contract.nonce()).toNumber()
```
mikheevm, line 267: nitpick: I can see at some time in the future a feature request for a way to set a custom nonce. No need to address it now but better to keep it in mind when designing the project.

src/managers/moduleManager.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  async isModuleEnabled(moduleAddress: string): Promise<boolean> {
+    return this.#contract.isModuleEnabled(moduleAddress)
+  }
+
+  async encodeEnableModuleData(moduleAddress: string): Promise<string> {
+    const isValidAddress = this.#ethers.utils.isAddress(moduleAddress)
+    if (!isValidAddress || moduleAddress === zeroAddress || moduleAddress === SENTINEL_MODULES) {
+      throw new Error('Invalid module address provided')
+    }
```
mikheevm, line 26: Could be possibly extracted to a function for reusability and readability (as it will have a meaningful name)

mikheevm: It's also used in ownerManager

mikheevm (approved):

src/managers/types.ts
germartinez: I forgot to remove them. They are not used with the Specific methods for each contract call.

src/EthersSafe.ts
germartinez: This was implemented in the other PR with some improvements. (Not pushed yet) :)

PR #23, opened 2021-05-06
MultiSend

germartinez:


README.md
```diff
@@ -136,12 +161,12 @@ Returns the connected signer.
...
 ```
 
-### getAddress
+### getSafeAddress
 
 Returns the address of the current Safe Proxy contract.
 
 ```js
-const address = safeSdk.getAddress()
+const address = safeSdk.getSafeAddress()
```
rmeissner, line 169: This feels a little weird as the `safeSdk` is an `EthersSafe` (so in the future maybe even only a `Safe` when we have web3 and ethers support) and calling `getSafeAddress` on that feels like a duplicate. I would leave it at `getAddress`, and just be explicit for the multiSend address

README.md
```diff
@@ -236,22 +261,27 @@ Each of the transactions provided as input to this function must be an object wi
...
+  data: '0x<data>',
+  value: '<eth_value_in_wei>'
+}
+const partialTx2: SafeTransactionDataPartial = {
   to: '0x<address>',
   data: '0x<data>',
   value: '<eth_value_in_wei>'
 }
-const safeTransaction = await safeSdk.createTransaction(partialTx)
+const safeTransaction = await safeSdk.createTransaction([partialTx1, partialTx2])
```
rmeissner, line 274: could we use the `varargs` format here?

src/configuration/contracts.ts
```diff
@@ -0,0 +1,38 @@
+interface ContractNetworkConfig {
```
rmeissner, line 1: you could pull this information from https://github.com/gnosis/safe-deployments

src/utils/transactions/utils.ts
```diff
@@ -32,3 +50,16 @@ export async function standardizeSafeTransaction(
...
+const encodeMetaTransaction = (tx: SafeBasicTransactionData): string => {
+  const data = utils.arrayify(tx.data)
+  const encoded = utils.solidityPack(
+    ['uint8', 'address', 'uint256', 'uint256', 'bytes'],
+    [tx.operation, tx.to, tx.value, data.length, data]
+  )
+  return encoded.slice(2)
+}
+
+export const encodeMultiSendData = (txs: SafeBasicTransactionData[]): string => {
```
rmeissner, line 63: Do we have tests for this method`?

src/utils/transactions/SafeTransaction.ts
```diff
@@ -5,10 +5,14 @@ export enum OperationType {
   DelegateCall // 1
 }
 
-export interface SafeTransactionData {
+export interface SafeBasicTransactionData {
```
rmeissner, line 8: No super strong opinion but I would probably prefer a naming like:

`SafeTransactionData` or `MetaTransactionData` (for the base)
and `MultisigSafeTransactionData` or similar ... but again ... probably not worth spending too much time on as none of them feel perfect :P

README.md
```diff
@@ -97,27 +100,49 @@ All the signatures used to execute the transaction are available at `tx.signatur
 Returns an instance of the Safe Core SDK with the `providerOrSigner` connected to the `safeAddress`.
 
 ```js
-const safeSdk = await EthersSafe.create(ethers, safeAddress, providerOrSigner)
+const safeSdk = await EthersSafe.create({ ethers, safeAddress, providerOrSigner })
```
rmeissner, line 103: Making the params into an object is a breaking change, right? 

This is not a problem (as we are at a 0.x version), just wondering ;)

rmeissner (commented):
I would love to see some tests for the multisend encoding else there is nothing blocking

src/configuration/contracts.ts
germartinez: Much better :)

src/utils/transactions/utils.ts
germartinez: There is this one (not a unit test of the method), but I will add a contract call too: https://github.com/gnosis/safe-core-sdk/pull/23/files#diff-c2f7973a9f5098cdc60417d715910ce586a7eae4a43b63cd65935b545ca99791R214

src/utils/transactions/SafeTransaction.ts
germartinez: Yes, I wasn't convinced with the name I used either.
As you said, I think I like `MultisigSafeTransactionData` and `MultisigSafeTransactionDataPartial` for the types that contain all the params (required and optional respectively) and `MetaTransactionData` for the one that only contains the parameters needed in the transactions inside the MultiSend.

germartinez:
Thanks @rmeissner!

src/configuration/contracts.ts
germartinez: TBD in a separate PR

src/utils/transactions/utils.ts
germartinez: Added the multisend with contract calls here: https://github.com/gnosis/safe-core-sdk/pull/23/files#diff-5535ed79ff3f49019cd635181eeed03ad747f8fb0d63b014008287dba9e8fe77R249

.github/workflows/test.yml
```diff
@@ -13,4 +13,4 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn install
-      - run: yarn test
+      - run: yarn sdk:test
```
mikheevm, line 16: should it run tests in all packages?

packages/safe-core-sdk/src/EthersSafe.ts
```diff
@@ -27,14 +33,18 @@ class EthersSafe implements Safe {
...
-  ): Promise<EthersSafe> {
+  static async create({
+    ethers,
+    safeAddress,
+    providerOrSigner,
+    contractNetworks
+  }: EthersSafeConfig): Promise<EthersSafe> {
     const safeSdk = new EthersSafe()
-    await safeSdk.init(ethers, safeAddress, providerOrSigner)
+    await safeSdk.init({ ethers, safeAddress, providerOrSigner, contractNetworks })
```
mikheevm, line 47: `contractNetworks` is missing in jsdoc definition

packages/safe-core-sdk/src/EthersSafe.ts
```diff
@@ -45,13 +55,16 @@ class EthersSafe implements Safe {
...
-  private async init(
-    ethers: any,
-    safeAddress: string,
-    providerOrSigner?: Provider | Signer
-  ): Promise<void> {
+  private async init({
+    ethers,
+    safeAddress,
+    providerOrSigner,
+    contractNetworks
```
mikheevm, line 66: same here for `contractNetworks`

packages/safe-core-sdk/src/EthersSafe.ts
```diff
@@ -63,28 +76,40 @@ class EthersSafe implements Safe {
...
-  async connect(providerOrSigner: Provider | Signer, safeAddress?: string): Promise<EthersSafe> {
-    return await EthersSafe.create(
-      this.#ethers,
-      safeAddress || this.#safeContract.address,
-      providerOrSigner
-    )
+  async connect({
+    providerOrSigner,
+    safeAddress,
+    contractNetworks
```
mikheevm, line 105: same here for `contractNetworks`

packages/safe-core-sdk/src/managers/contractManager.ts
```diff
@@ -0,0 +1,72 @@
...
+      throw new Error('MultiSend contract is not deployed in the current network')
+    }
+    this.#multiSendContract = new ethers.Contract(
+      contracts.multiSendAddress,
+      MultiSendAbi,
+      currentProviderOrSigner
+    )
+  }
+
+  get safeContract() {
```
mikheevm, line 63: i'm not an oop expert but this getter is the same as just accessing the property, why make it private?

packages/safe-core-sdk/src/managers/ownerManager.ts
```diff
@@ -53,7 +53,8 @@ class OwnerManager {
   }
 
   async getOwners(): Promise<string[]> {
-    return this.#safeContract.getOwners()
+    const owners = await this.#safeContract.getOwners()
+    return [...owners]
```
mikheevm, line 57: why does it have to be this way?

packages/safe-core-sdk/src/utils/transactions/gas.ts
```diff
@@ -15,24 +15,24 @@ function estimateDataGasCosts(data: any): number {
   return data.match(/.{2}/g).reduce(reducer, 0)
 }
 
-export async function estimateTxGas(
-  contract: GnosisSafe,
+export const estimateTxGas = async (
```
mikheevm, line 18: this was changed to be an arrow function but the function `estimateGasForTransactionExecution` below and `estimateDataGasCosts` above is defined with `function` keyword, I think it's better to stay consistent

packages/safe-core-sdk/src/utils/transactions/gas.ts
```diff
@@ -1,6 +1,6 @@
 import { BigNumber } from 'ethers'
 import { GnosisSafe } from '../../../typechain'
-import SafeTransaction from './SafeTransaction'
+import SafeTransaction, { OperationType } from './SafeTransaction'
 
 function estimateDataGasCosts(data: any): number {
```
mikheevm, line 5: This function won't work for anything other than a string (or anything that defines `.match`), why does it use any type for its param?

packages/safe-core-sdk/src/utils/transactions/utils.ts
```diff
@@ -0,0 +1,65 @@
...
+const encodeMetaTransaction = (tx: MetaTransactionData): string => {
+  const data = utils.arrayify(tx.data)
+  const encoded = utils.solidityPack(
+    ['uint8', 'address', 'uint256', 'uint256', 'bytes'],
+    [tx.operation, tx.to, tx.value, data.length, data]
+  )
+  return encoded.slice(2)
+}
+
+export const encodeMultiSendData = (txs: MetaTransactionData[]): string => {
```
mikheevm, line 63: same nitpick here - functions here mixed with arrow/function definitions

tsconfig.base.json
```diff
@@ -18,14 +17,11 @@
...
-    "baseUrl": "src", /* Base directory to resolve non-absolute module names. */
     "allowSyntheticDefaultImports": true, /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
     "esModuleInterop": true, /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
     "experimentalDecorators": true, /* Enables experimental support for ES7 decorators. */
     "skipLibCheck": true, /* Skip type checking of declaration files. */
     "forceConsistentCasingInFileNames": true, /* Disallow inconsistently-cased references to the same file. */
-    "resolveJsonModule": true,
-  },
-  "include": ["src"],
-  "exclude": ["dist", "node_modules", "coverage"]
```
mikheevm, line 30: I think you could leave `exclude` there

packages/safe-core-sdk/tests/utils/setup.ts
```diff
@@ -0,0 +1,72 @@
...
+    '0x',
+    AddressZero,
+    AddressZero,
+    0,
+    AddressZero
+  )
+  return template
+}
+
+export const getMultiSend = async (): Promise<MultiSend> => {
```
mikheevm, line 50: how does this file compare to `tests/utils/setupContracts.ts` seems they're the same

packages/safe-core-sdk/tests/threshold.test.ts
```diff
@@ -1,35 +1,57 @@
...
+import { deployments, ethers, waffle } from 'hardhat'
 import EthersSafe from '../src'
+import { ContractNetworksConfig } from '../src/configuration/contracts'
+import { GnosisSafe } from '../typechain'
+import { getMultiSend, getSafeWithOwners } from './utils/setup'
 import { getAccounts } from './utils/setupConfig'
-import { getSafeWithOwners } from './utils/setupContracts'
 chai.use(chaiAsPromised)
 
+interface SetupTestsResult {
```
mikheevm, line 11: This interface seems to be defined almost in each test file, maybe move it out to `utils`?

packages/safe-core-sdk/src/managers/ownerManager.ts
germartinez: Otherwise it returns a frozen array.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
`"As an object, an array can be frozen; after doing so, its elements cannot be altered and no elements can be added to or removed from the array."`

mikheevm: but it doesn't matter in our case, does it?

germartinez: Not in our case, but I think it will avoid some confusion for many developers if they try to sort the array or modify it and it does not work.

packages/safe-core-sdk/src/managers/contractManager.ts
germartinez: to prevent from being modified

mikheevm: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_instance_fields

TIL you also can't access them otherwise. Btw, in safe-apps-sdk we don't use it anymore because of https://github.com/gnosis/safe-apps-sdk/issues/141

mikheevm (approved):

PR #26, opened 2021-05-13
Fix signMessage and test against ganache

germartinez:
- After generating a signature with `signMessage` it checks if the message had a prefix or not to adjust the `v` value accordingly
- Add script `yarn test:ganache` to run the tests against ganache optionally
- Refactor tests to work with acounts from ganache or hardhat
- Update README file


src/utils/signatures/index.ts
```diff
@@ -9,3 +12,57 @@ export function generatePreValidatedSignature(ownerAddress: string): SafeSignatu
...
+    const recoveredData = ecrecover(
+      Buffer.from(txHash.slice(2), 'hex'),
+      rsvSig.v,
+      rsvSig.r,
+      rsvSig.s
+    )
+    const recoveredAddress = bufferToHex(pubToAddress(recoveredData))
+    hasPrefix = !sameString(recoveredAddress, ownerAddress)
+  } catch (e) {
+    hasPrefix = true
```
rmeissner, line 65: Wouldn't it make more sense to throw? E.g. if we cannot recover it, something is probably wrong

src/utils/signatures/index.ts
```diff
@@ -9,3 +12,57 @@ export function generatePreValidatedSignature(ownerAddress: string): SafeSignatu
...
+): boolean {
+  let hasPrefix
+  try {
+    const rsvSig = {
+      r: Buffer.from(signature.slice(2, 66), 'hex'),
+      s: Buffer.from(signature.slice(66, 130), 'hex'),
+      v: parseInt(signature.slice(130, 132), 16)
+    }
+    const recoveredData = ecrecover(
+      Buffer.from(txHash.slice(2), 'hex'),
```
rmeissner, line 57: Should we check this also for the prefixed hash and then throw if it matches neither? In the future we could this way also properly handle contract signatures

package.json
```diff
@@ -11,6 +11,7 @@
   ],
   "scripts": {
     "build": "yarn rimraf dist && hardhat compile && tsc",
+    "test:ganache": "export TEST_NETWORK=ganache && hardhat --network localhost deploy && nyc hardhat --network localhost test",
```
rmeissner, line 14: Should we run this in the CI?

rmeissner (approved):
Looks like a good base to further improve the handling of the signatures in the future.

E.g. check also if it is a contract signature (would be part of EIP-1271 issue)

tests/utils/setupConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ethers, waffle, Web3 } from 'hardhat'
+
+interface Account {
+  signer: Signer
+  address: string
+}
+
+async function getGanacheAccounts(): Promise<Account[]> {
+  const web3 = new Web3('http://localhost:8545')
+  const provider = new ethers.providers.Web3Provider(web3.currentProvider as any)
```
mikheevm, line 11: why any?

tests/utils/setupConfig.ts
```diff
@@ -0,0 +1,36 @@
...
+  const accounts: Account[] = []
+  for (let i = 0; i < 10; i++) {
+    const signer: Signer = provider.getSigner(i)
```
mikheevm, line 14: aren't types inherited?

mikheevm (approved):

PR #31, opened 2021-06-03
Add Safe Service Client

germartinez:


geigerzaehler:
This is much appreciated!

While trying this out in [Radicle Upstream](https://github.com/radicle-dev/radicle-upstream) we realized that `getPendingTransactions(safeAddress)` for a safe that does not exist yet returns an object that does not match the return type. It returns the object `{ nonce_Gte: ["Enter a number"] }`. Now that we know that we can react to it but having incorrect types made this a bit confusing. In general it seems that error responses are just returned as-is. It would be great if these were reflected in the return type or would instead throw an exception :bowing_man:

packages/safe-service-client/README.md
```diff
@@ -0,0 +1,233 @@
...
+```js
+const collectibles: SafeCollectibleResponse[] = await safeService.getCollectibles(safeAddress)
+```
+
+### getTokens
+
+Returns the list of all the ERC20 tokens handled by the Safe.
+
+```js
+const tokens: TokenInfoListResponse = await safeService.getTokens()
```
mikheevm, line 216: Is there `safeAddress` arg missing? 

EDIT: Ok, it seems to be a token list. Maybe better to rename it to getTokenList and update description

packages/safe-service-client/src/SafeTransactionService.ts
```diff
@@ -0,0 +1,78 @@
...
+  // Transactions
+  getSafeCreationInfo(safeAddress: string): Promise<SafeCreationInfoResponse>
+  estimateSafeTransaction(
+    safeAddress: string,
+    safeTransaction: SafeMultisigTransactionEstimate
+  ): Promise<SafeMultisigTransactionEstimateResponse>
+  proposeTransaction(
+    safeAddress: string,
+    transaction: SafeTransactionData,
+    transactionHash: string,
```
mikheevm, line 57: I was confused by transactionHash (what tx hash is it?) naming, maybe it's better to use `contractTransactionHash` as it's named in the request body or `safeTxHash`

packages/safe-service-client/src/utils/httpRequests.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+interface HttpRequest {
+  url: string
+  method: HttpMethod
+  body?: Object
+}
+
+export async function sendRequest<T>({ url, method, body }: HttpRequest): Promise<T> {
+  let response: AxiosResponse<any>
```
mikheevm, line 16: You can pass `T` here too, it takes an argument for the response data

mikheevm (approved):
👏

viraj124:
@germartinez what's the eta of this merge?
 since I was facing this problem
`safe core sdk has a execute transaction method which requires an object of SafeTransaction interface, but on the @gnosis.pm/safe-core-sdk & @gnosis.pm/safe-service-client there is no read method that returns the  SafeTransaction object by providing the txHash`
looks like this or solves that issue by the return type matching what execute transaction  method needs

viraj124:
basically, I am looking for a read method that returns an object of SafeTransaction interface

PR #37, opened 2021-06-07
Handle WalletConnect signatures

germartinez:
Add the right prefix to the signature when signing with an owner that is connected via WalletConnect


packages/safe-core-sdk/src/utils/signatures/index.ts
```diff
@@ -13,32 +13,17 @@ export function generatePreValidatedSignature(ownerAddress: string): SafeSignatu
...
-      throw new Error('Invalid signature')
-  }
-  signature = signature.slice(0, -2) + signatureV.toString(16)
-  return new EthSignSignature(signerAddress, signature)
+function isWalletConnectProvider(signer: Signer) {
+  return !!((signer?.provider as any)?.provider?.wc)
+}
+
+async function generateWalletConnectSignature(ethers: any, hash: any, signer: Signer): Promise<string> {
+  const prefix = ethers.utils.toUtf8Bytes(`\x19Ethereum Signed Message:\n${hash.length}`)
```
rmeissner, line 21: Doesn't this depend on the connected wallet? With which wallets did you test it?

germartinez: I tested with MetaMask directly and with MetaMask mobile via WalletConnect

rmeissner: I would recommend to test also with other wallets, as MetaMask doesn't follow the standard when it comes to `eth_sign`. But this should not be an issues since the signatures is also valid without the prefix, but in this case you don't need to add `+4` to `v` ... So I would prefer if we don't add the prefix manually and rather check if it was used and adjust `v` according to that

mikheevm (approved):

PR #38, opened 2021-06-14
Add method createRejectionTransaction

germartinez:
Adds the method rejectTransaction that invalidates a pending Safe transaction.


packages/safe-core-sdk/src/EthersSafe.ts
```diff
@@ -255,6 +255,22 @@ class EthersSafe implements Safe {
     return new SafeTransaction(standardizedTransaction)
   }
 
+  /**
+   * Returns a Safe transaction ready to be signed by the owners that invalidates a pending Safe transaction.
+   *
+   * @param safeTransaction - The pending Safe transaction to be invalidated
+   * @returns The Safe transaction that invalidates the pending Safe transaction
+   */
+  async rejectTransaction(safeTransaction: SafeTransaction): Promise<SafeTransaction> {
```
mikheevm, line 264: I wonder if the name should be `createRejectionTransaction`. This fits more to the existing `createTransaction` and in my opinion closer to what is happening.

packages/safe-core-sdk/tests/execution.test.ts
```diff
@@ -197,21 +212,60 @@ describe('Transactions execution', () => {
         .expect(safeInitialBalance.toString())
         .to.be.eq(safeFinalBalance.add(BigNumber.from(tx.data.value).toString()))
     })
+  
+    it('should fail if a transaction that was rejected is executed', async () => {
```
mikheevm, line 216: I found this test description hard to understand, I think something like `should fail if you try to execute rejected transaction`. Maybe I'm nitpicking, but I hard to read the test to understand what's happening

packages/safe-core-sdk/src/EthersSafe.ts
mikheevm: Maybe in the SDK, we'd also want API to be about replacing a transaction. In the UI it was simplified to "cancelation", but maybe in the SDK, we can be more realistic about what's happening behind the scenes. The API can be way simpler too: `.replaceTransaction(nonce)`

germartinez: I like `createRejectionTransaction(nonce)`. Using the word "replace" is not my favourite option because it is true that we are replacing a transaction, but always with an empty transaction, not a custom one.

packages/safe-core-sdk/tests/execution.test.ts
germartinez: What do you think about `should fail if the user tries to execute a transaction that was invalidated`?

mikheevm: I don't like it because we don't really use `invalidated`, if you replace it with `rejected` then it sounds good to me. The description I suggested is even simpler IMO because it doesn't contain [passive voice](https://en.wikipedia.org/wiki/Passive_voice)

rmeissner (approved):

PR #44, opened 2021-06-28
Allow to override transaction options

germartinez:
Adds an `options` parameter to `executeTransaction()` in order to override the `gasLimit` or set a `gasPrice`.

See https://github.com/gnosis/safe-core-sdk/issues/43


ilanolkies:
Thanks! LRGTM

mikheevm (approved):

PR #50, opened 2021-07-20
Reduce package size

germartinez:
- Include only required Typechain files
- Improve imports


packages/safe-core-sdk/scripts/generateTypechainFiles.ts
```diff
@@ -0,0 +1,79 @@
+const { execSync } = require('child_process')
```
mikheevm, line 1: I don't really get why are we optimizing this, this won't end up in the customer's production bundle because these are type definitions. What are we saving? 1 second from type generating or few kilobytes of a developer bandwidth?

mikheevm: What it optimizes is not worth it compared to what complexity it brings

mikheevm (approved):

PR #61, opened 2021-09-06
(safe-core-sdk) Add nonce to Safe config methods

germartinez:


mikheevm:
Do you think it makes sense to take `opts: TransactionDataPartial` instead of a single nonce like `createTransaction`? I can imagine rare use cases where this is needed: for example, using meta transactions.

https://github.com/gnosis/safe-core-sdk/blob/39982001593097892fc614252564293c01988e34/packages/safe-core-sdk-types/src/types.ts#L23

packages/safe-core-sdk/src/Safe.ts
```diff
@@ -353,75 +383,92 @@ class Safe {
...
    * @returns The Safe transaction ready to be signed
    * @throws "Invalid new owner address provided"
    * @throws "Invalid old owner address provided"
    * @throws "New address provided is already an owner"
    * @throws "Old address provided is not an owner"
    */
-  async getSwapOwnerTx(oldOwnerAddress: string, newOwnerAddress: string): Promise<SafeTransaction> {
+  async getSwapOwnerTx(
+    { oldOwnerAddress, newOwnerAddress }: SwapOwnerTxParams,
+    options?: SafeTransactionOptionalProps
```
mikheevm, line 465: Do you think it makes sense to have a separate interface/type for options excluding the operation? For example, for the owner management methods, you don't want to change the operation
```
export type CallTransactinoOptions = Omit<SafeTransactionOptionalProps, 'operation'>
```

germartinez: Makes sense. I will update that

mikheevm (approved):
🎉

PR #62, opened 2021-09-06
(safe-service-client) Add option to exclude/include spam tokens

germartinez:


packages/safe-service-client/src/SafeServiceClient.ts
```diff
@@ -339,9 +339,9 @@ class SafeServiceClient implements SafeTransactionService {
    * @throws "404 Safe not found"
    * @throws "422 Safe address checksum not valid"
    */
-  async getBalances(safeAddress: string): Promise<SafeBalanceResponse[]> {
+  async getBalances(safeAddress: string, excludeSpamTokens = true): Promise<SafeBalanceResponse[]> {
```
mikheevm, line 342: Should it be an opts object instead so we could easily add other options in the future?

mikheevm (approved):

PR #64, opened 2021-09-14
(safe-service-client) Handle error messages and add e2e tests

germartinez:
## Safe Service Client
- Handle error messages
- Add e2e tests


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,25 @@
...
+        node-version: [14.x]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: yarn install
+      - run: yarn build
+      - run: yarn test:ci
```
mikheevm, line 25:
```suggestion
      - run: |
           yarn install
           yarn build
           yarn test:ci
```

nitpick, but I find this way nicer

mikheevm (approved):
👏

.github/workflows/e2e-test.yml
germartinez: True, thanks!

PR #65, opened 2021-09-20
Throw exception if there are not enough funds

germartinez:
Throws the exception `Not enough Ether funds` when the Safe Ether balance is lower than the value of a tx that will be executed

### DIFF
--- a/packages/safe-core-sdk/src/Safe.ts
+++ b/packages/safe-core-sdk/src/Safe.ts
@@ -529,6 +529,14 @@ class Safe {
       )
     }
 
+    const value = BigNumber.from(safeTransaction.data.value)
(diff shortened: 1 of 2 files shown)


packages/safe-core-sdk/src/Safe.ts
```diff
@@ -529,6 +529,14 @@ class Safe {
       )
     }
 
+    const value = BigNumber.from(safeTransaction.data.value)
+    if (value.gt(BigNumber.from(0))) {
```
mikheevm, line 533:
```suggestion
    if (!value.isZero()) {
```
according to https://docs.ethers.io/v5/api/utils/bignumber/ such method is available

mikheevm (approved):

PR #70, opened 2021-09-23
(safe-service-client) Remove axios

germartinez:


mikheevm (approved):
the power of abstraction at its best right here

PR #76, opened 2021-09-28
Set optional params to MultiSend transactions

### LINKED ISSUE #73: Set custom/optional parameters in MultiSend transaction
## Description

When creating a multisend transaction the user is not allowed to set the optional params included in `SafeTransactionDataPartial` (`safeTxGas`, `baseGas`, `gasPrice`, `gasToken`, `refundReceiver` and `nonce`).

## Environment

- Safe Core SDK last version

## Expected result

The user is allowed to set the optional params included in `SafeTransactionDataPartial` (`safeTxGas`, `baseGas`, `gasPrice`, `gasToken`, `refundReceiver` and `nonce`) when creating a multisend transaction.

germartinez:
Read more about MultiSend transactions: https://docs.gnosis.io/safe/docs/contracts_details/#multisendsol

## What it solves
Resolves #73

## How this PR fixes it
It adds the param `options` to the `craeteTransaction` method when the `transactions` param is an array of `MetaTransactionData` objects. If the param `transactions` is a single `SafeTransactionDataPartial` object, the optional params are included in that type.

```
function createTransaction(transactions: SafeTransactionDataPartial)
function createTransaction(transactions: MetaTransactionData[], options?: SafeTransactionDataPartial)
```

### Example:

If we want to set a specific `nonce` for a single transaction, the property `nonce` is inside the type `SafeTransactionDataPartial` and should be set like this:
```js
const singleTransaction: SafeTransactionDataPartial = {
  ...,
  nonce: 5
}
const safeTx = sdk.createTransaction(singleTransaction)
```

If we want to set a specific `nonce` for a multisend transaction, we use the param `options` with that `nonce`:
```js
const multisendTransaction: MetaTransactionData[] = [
  {
    to: '...',
    value: '...',
    data: '...'
  },
  { ... }
]
const options: SafeTransactionOptionalProps = {
   ...,
  nonce: 5
}
const safeTx = sdk.createTransaction(multisendTransaction, options)
```


iamacook (commented):
There seems to be some similarities between the optional types and the types from `safe-core-sdk-types`. If there is crossover, maybe you can `Pick<Type, ...>` from them so you need to worry if they change?

Apart from that, all looks good to me :+1:

iamacook (approved):

PR #78, opened 2021-09-29
Make signatures nullable in SafeMultisigTransactionResponse

germartinez:
Makes the property `signatures` nullable in the type `SafeMultisigTransactionResponse` returned by the Safe Transaction Service.

### DIFF
--- a/packages/safe-service-client/src/types/safeTransactionServiceTypes.ts
+++ b/packages/safe-service-client/src/types/safeTransactionServiceTypes.ts
@@ -134,7 +134,7 @@ export type SafeMultisigTransactionResponse = {
   readonly dataDecoded?: string
   readonly confirmationsRequired: number
   readonly confirmations?: SafeMultisigConfirmationResponse[]
-  readonly signatures: string
+  readonly signatures?: string
 }
 
 export type SafeMultisigTransactionListResponse = {
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,5 @@ deployments
 
 # Typechain
 typechain
+
+openapi/
--- a/packages/safe-service-client/scripts/generateTxServiceTypes.sh
+++ b/packages/safe-service-client/scripts/generateTxServiceTypes.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p openapi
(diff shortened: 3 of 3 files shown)


packages/safe-service-client/src/types/safeTransactionServiceTypes.ts
```diff
@@ -134,7 +134,7 @@ export type SafeMultisigTransactionResponse = {
   readonly dataDecoded?: string
   readonly confirmationsRequired: number
   readonly confirmations?: SafeMultisigConfirmationResponse[]
-  readonly signatures: string
+  readonly signatures?: string
```
anxolin, line 137: The endpoint can return `null` which is different than `undefined` 

image

If you want to reflect the returned data type should be `signatures: string | null`

```suggestion
  readonly signatures: string | null
```

anxolin (commented):
Thanks for doing the change

packages/safe-service-client/src/types/safeTransactionServiceTypes.ts
anxolin: @Uxio0 I believe the issue is also in swagger docs, right? 

https://safe-transaction.rinkeby.gnosis.io/

image

anxolin: Should it be nullable? https://swagger.io/docs/specification/data-models/data-types/#null ?

anxolin: Probably it would be good to revisit the swagger docs for the other fields, cause at least `transactionHash` is also null while the transaction is not mined

germartinez: Thanks!

I talked with Uxío and he will make the field not required.

I will also generate all the types with [openapi-typescript](https://www.npmjs.com/package/openapi-typescript) so I will wait until this is updated: https://safe-transaction.rinkeby.gnosis.io/?format=openapi

Uxio0: Yes, there's a typo on the docs, thanks for reporting

PR #82, opened 2021-10-04
Fix path in scripts

germartinez:
Fix path in scripts.
It should be `examples/*` instead of `example/*`

e2e tests have been updated after some changes were made using the Safe included in the tests.


iamacook (approved):
👍

PR #83, opened 2021-10-05
Add a PR template and typedoc site generation

katspaugh:
This PR adds a PR template to make linking PRs to issues easier.

It also introduces a github action that generates HTML docs based on types and deploys them to GitHub Pages.
Once this is merged, GitHub Pages need to be enabled.

## Docs screenshot
Screenshot 2021-10-05 at 13 07 19

## How to enable GitHub pages
Screenshot 2021-10-05 at 13 09 15

PR #84, opened 2021-10-05
Fix the typedoc action

katspaugh:
## What it solves
After merging #83, it turned out the build isn't working in the typedoc action.
This PR is to debug and fix that.


katspaugh:
It works now 😄 
https://gnosis.github.io/safe-core-sdk/classes/default.html

germartinez (approved):

PR #98, opened 2021-11-05
Fix code example mistake

### LINKED ISSUE #97: Error in documentation of safe-core-sdk
## Description

The documentation of the `safe-core-sdk` has an error in the `API Reference` / `connect` section:
https://github.com/gnosis/safe-core-sdk/tree/main/packages/safe-core-sdk#api-reference

The code sample for the `contractNetworks` property was probably copy-pasted from the `create` section. I assume it should use `connect` instead of `create`.

## Expected result

Documentation should read like:

```
const safeSdk = await Safe.connect({ ethAdapter, safeAddress, contractNetworks })
```

cmdzro:
## What it solves
Resolves #97 

## How this PR fixes it
Replacing `create` with `connect`.


cmdzro:

PR #106, opened 2021-11-17
Multisend fix

### LINKED ISSUE #101: Update condition when creating a transaction: single vs batch
## Description
Update the condition inside `createTransaction(...)` that decides if the SDK will send a single or a batch transaction.
When passing an array of transaction as a parameter, take into account:
 - If the length of the array is 0
 - If the length of the array is 1 -> create single transaction

Check: https://gist.github.com/cmdzro/0547725ca6faa2067e7edf3c70b697f1#gistcomment-3954987

germartinez:
## What it solves
Resolves #101 

## How this PR fixes it
- Considers that transaction arrays with length = 1 are single transactions, not Multisend.
- Does not accept transaction arrays with length = 0
- Tests are updated to consider these use cases

### DIFF
--- a/packages/safe-core-sdk/src/Safe.ts
+++ b/packages/safe-core-sdk/src/Safe.ts
@@ -11,7 +11,7 @@ import ContractManager from './managers/contractManager'
 import ModuleManager from './managers/moduleManager'
 import OwnerManager from './managers/ownerManager'
 import { ContractNetworksConfig } from './types'
-import { sameString } from './utils'
+import { isMetaTransactionArray, sameString } from './utils'
 import { generatePreValidatedSignature, generateSignature } from './utils/signatures'
 import { estimateGasForTransactionExecution } from './utils/transactions/gas'
 import EthSafeTransaction from './utils/transactions/SafeTransaction'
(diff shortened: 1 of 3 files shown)


packages/safe-core-sdk/src/Safe.ts
```diff
@@ -260,11 +261,17 @@ class Safe {
   ): Promise<SafeTransaction>
   async createTransaction(
     safeTransactions: SafeTransactionDataPartial | MetaTransactionData[],
-    options?: SafeTransactionDataPartial
+    options?: SafeTransactionOptionalProps
   ): Promise<SafeTransaction> {
-    if (safeTransactions instanceof Array) {
+    const isTxArray = Array.isArray(safeTransactions)
+    if (isTxArray && (safeTransactions as MetaTransactionData[]).length === 0) {
```
iamacook, line 267: Will this always be `MetaTransactionsData[]`?

iamacook: As well as the other casts you make in this function?

iamacook: You could also just cast it once under a new variable, no?

iamacook (commented):
Looking good, just one small thing

packages/safe-core-sdk/src/Safe.ts
germartinez: It can be `MetaTransactionsData[]` or `SafeTransactionDataPartial`

iamacook: Maybe make a type checking function that returns `value is X`?

iamacook (approved):
🚀

PR #121, opened 2021-12-03
Set safeTxGas=0 when safeVersion>=1.3.0

### LINKED ISSUE #100: Send 0 for safeTxGas
## Overview
`safeTxGas` is a workaround that was used so RPC providers could estimate correctly the gas used by a Safe transaction. This way we could get a valid `gasLimit` and avoid running out of gas because transaction was estimated lower than the real cost.
Since v1.3.0 this workaround is not necessary any more as now a transaction failing by low gas will revert instead of succeed, and we could simply send a 0 value and let the wallets to estimate the full cost in gas of the transaction.

From the contract code comments:
```
If no safeTxGas and no gasPrice was set (e.g. both are 0),
then the internal tx is required to be successful
This makes it possible to use `estimateGas` without issues,
as it searches for the minimum gas where the tx doesn't revert
```

## Goals
* Send `safeTxGas=0` when sending a transaction from a Safe that is using at least v1.3.0 version of the smart contract
* Export `standardizeSafeTransactionData`

## Requirements
* For v1.3.0 safes, send a `safeTxGas=0`
* For earlier version (esp. v1.1.1), keep estimating and sending the value

germartinez:
## What it solves
Resolves #100

### DIFF
--- a/packages/safe-core-sdk/src/utils/transactions/utils.ts
+++ b/packages/safe-core-sdk/src/utils/transactions/utils.ts
@@ -6,6 +6,7 @@ import {
   SafeTransactionData,
   SafeTransactionDataPartial
 } from '@gnosis.pm/safe-core-sdk-types'
+import semverSatisfies from 'semver/functions/satisfies'
 import GnosisSafeContract from '../../contracts/GnosisSafe/GnosisSafeContract'
 import EthAdapter from '../../ethereumLibs/EthAdapter'
 import { ZERO_ADDRESS } from '../constants'
@@ -37,9 +38,12 @@ export async function standardizeSafeTransactionData(
     refundReceiver: tx.refundReceiver || ZERO_ADDRESS,
     nonce: tx.nonce ?? (await safeContract.getNonce())
   }
-  const safeTxGas =
-    tx.safeTxGas ??
-    (await estimateTxGas(
+  let safeTxGas: number
+  const safeVersion = await safeContract.getVersion()
+  if (semverSatisfies(safeVersion, '>=1.3.0')) {
+    safeTxGas = 0
+  } else {
+    safeTxGas = (tx.safeTxGas) ?? (await estimateTxGas(
       safeContract,
       ethAdapter,
       standardizedTxs.to,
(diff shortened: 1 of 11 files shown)


packages/safe-core-sdk/src/utils/transactions/utils.ts
```diff
@@ -37,16 +38,22 @@ export async function standardizeSafeTransactionData(
...
-      safeContract,
-      ethAdapter,
-      standardizedTxs.to,
-      standardizedTxs.value,
-      standardizedTxs.data,
-      standardizedTxs.operation
-    ))
+  let safeTxGas: number
+  const safeVersion = await safeContract.getVersion()
+  if (semverSatisfies(safeVersion, '>=1.3.0')) {
```
katspaugh, line 43: Maybe put `'>=1.3.0'` in some sort of config, [similar to safe-react](https://github.com/gnosis/safe-react/blob/a2cd631b8f765c10a1fbf55de9290121e5da0e9b/src/logic/safe/utils/safeVersion.ts)?

iamacook, line 43: Maybe it makes sense to create a lookup for the versions?

packages/safe-service-client/tests/endpoint.test.ts
```diff
@@ -33,210 +33,14 @@ describe('Endpoint tests', () => {
...
       const safeTxData: SafeTransactionDataPartial = {
         to: '0xa33d2495760462018275994d85117600bd58221e',
         data: '0x',
         value: '123456789',
         operation: 1,
-        safeTxGas: 123,
+        safeTxGas: 0,
         baseGas: 0,
         gasPrice: 0,
         gasToken: '0x0000000000000000000000000000000000000000',
```
iamacook, line 46: Do you use the a `ZERO_ADDRESS` often? Might be worthwhile creating a constant for it.

iamacook (approved):
Looking good! I just have a couple of suggestions

DiogoSoaress (commented):
I would just address the suggestion of using some kind of lookup for the `safe version`

packages/safe-service-client/tests/endpoint.test.ts
germartinez: It is only used twice in this test file

iamacook (approved):
🚀

PR #122, opened 2021-12-15
Update safe-deployments to v1.5.0

germartinez:
- Update safe-deployments to v1.5.0
- Update other dev dependencies


katspaugh (approved):
Thanks!

katspaugh:
Probably bump the version of this package as well?

PR #134, opened 2022-01-05
Fix contract calls in class GnosisSafeContract_V1_1_1

germartinez:
## What it solves
`super.contract` is undefined in classes `GnosisSafeContract_V1_1_1_Ethers` and `GnosisSafeContract_V1_1_1_Web3`

## How this PR fixes it
Call `this.contract`


iamacook (commented):
~~It might be safer to create a type check function instead of casting?~~

Sorry, I read it backwards 🤦‍♂️

iamacook (approved):
🚀

PR #136, opened 2022-01-05
feat: Add ens lookup functions

usame-algan:
## What it solves
Resolves [#3231](https://github.com/gnosis/safe-react/issues/3231)

## How this PR fixes it
Adds two new methods to the Web3Adapter for ens lookup and reverse lookup

### DIFF
--- a/packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
+++ b/packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
@@ -1,3 +1,5 @@
+import Web3 from 'web3'
+import { ethers } from 'ethers'
 import { BigNumber } from '@ethersproject/bignumber'
 import { SafeVersion } from '../contracts/config'
 import {
@@ -24,7 +26,7 @@ export interface Web3AdapterConfig {
 }
 
 class Web3Adapter implements EthAdapter {
-  #web3: any
+  #web3: Web3
   #signerAddress: string
 
   constructor({ web3, signerAddress }: Web3AdapterConfig) {
@@ -105,7 +107,7 @@ class Web3Adapter implements EthAdapter {
     return getGnosisSafeProxyFactoryContractInstance(safeVersion, proxyFactoryContract)
   }
 
-  getContract(address: string, abi: AbiItem[]): any {
+  getContract(address: string, abi: any): any {
     return new this.#web3.eth.Contract(abi, address)
   }
 
@@ -125,12 +127,23 @@ class Web3Adapter implements EthAdapter {
     return this.#web3.eth.sign(message, this.#signerAddress)
   }
 
-  estimateGas(transaction: EthAdapterTransaction, options?: string): Promise<number> {
+  estimateGas(transaction: EthAdapterTransaction, options?: any): Promise<number> {
     return this.#web3.eth.estimateGas(transaction, options)
   }
 
   call(transaction: EthAdapterTransaction): Promise<string> {
     return this.#web3.eth.call(transaction)
+  }
+
+  ensLookup(name: string): Promise<string> {
+    return this.#web3.eth.ens.getAddress(name)
+  }
+
+  async ensReverseLookup(address: string): Promise<string> {
+    const lookup = address.toLowerCase().substr(2) + '.addr.reverse'
+    const node = ethers.utils.namehash(lookup)
+    const ResolverContract = await this.#web3.eth.ens.getResolver(lookup);
+    return await ResolverContract.methods.name(node).call()
   }
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ deployments
 typechain
 
 openapi/
+.idea
--- a/packages/safe-core-sdk/tests/ens.test.ts
+++ b/packages/safe-core-sdk/tests/ens.test.ts
@@ -1,0 +1,41 @@
+import Web3 from 'web3'
+import { Web3Adapter } from '../src'
+import { ZERO_ADDRESS } from '../src/utils/constants'
+import chai from 'chai'
+
+describe.only("Web3Adapter", () => {
+  const web3 = new Web3(new Web3.providers.HttpProvider('https://rinkeby-light.eth.linkpool.io/', {}))
+  const web3Adapter = new Web3Adapter({web3, signerAddress: ZERO_ADDRESS})
+
+  describe("ENS direct lookup", () => {
+    it("returns an address for ens name", async () => {
+      const address = await web3Adapter.ensLookup("safe.eth")
+
+      chai.expect(address).match(/0x[0-9a-z]{40}/i)
+    })
+
+    it("throws an error for a non existing name", async () => {
+      try {
+        await web3Adapter.ensLookup("nonexistingname")
+      } catch(error) {
+        chai.expect((error as Error).message).to.equal("The resolver at 0x0000000000000000000000000000000000000000does not implement requested method: \"addr\".")
+      }
+    })
+  })
+
+  describe("ENS reverse lookup", () => {
+    it("returns a name for existing address", async () => {
+      const address = await web3Adapter.ensReverseLookup("0x203aBBf9F190009606C317439360c9f3c0874452")
+
+      chai.expect(address).to.equal("marco.eth")
+    })
+
+    it("throws an error if no name exists for address", async () => {
+      try {
+        await web3Adapter.ensReverseLookup("0x16B110D5b7583266B29159d89eF0d001adf6f6FD")
+      } catch(error: any) {
+        chai.expect(error.message).to.exist
+      }
+    })
+  })
+})
--- a/packages/safe-core-sdk/tsconfig.json
+++ b/packages/safe-core-sdk/tsconfig.json
@@ -5,5 +5,5 @@
     "outDir": "dist",
     "baseUrl": "src"
   },
-  "include": ["src", "src/**/*.json", "typechain/src"]
+  "include": ["src", "src/**/*.json", "typechain/src", "tests"]
 }


usame-algan:

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -24,7 +26,7 @@ export interface Web3AdapterConfig {
 }
 
 class Web3Adapter implements EthAdapter {
```
germartinez, line 28: Adding some context:
`EthAdapter` is an interface implemented by `Web3Adapter` and `EthersAdapter` classes where all the methods are defined in the interface and implemented using Web3 and ethers.js libraries in the 2 classes respectively. This way it doesn't matter if a Dapp uses ethers.js or web3, they will still be able to access the same functionality.

It would be cool if the 2 new methods are included in `EthAdapter` and `EthersAdapter`.

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -1,3 +1,5 @@
+import Web3 from 'web3'
+import { ethers } from 'ethers'
```
germartinez, line 2: ethers.js import should not be present in `Web3Adapter`

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -125,13 +127,24 @@ class Web3Adapter implements EthAdapter {
...
     return this.#web3.eth.call(transaction)
   }
+
+  ensLookup(name: string): Promise<string> {
+    return this.#web3.eth.ens.getAddress(name)
+  }
+
+  async ensReverseLookup(address: string): Promise<string> {
+    const lookup = address.toLowerCase().substr(2) + '.addr.reverse'
+    const node = ethers.utils.namehash(lookup)
```
germartinez, line 144: ethers.js library should not be used in `Web3Adapter` class

germartinez (changes requested):
:muscle: Cool feature!

lukasschor:
Nice! 😎

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
usame-algan: Unfortunately web3 doesn't provide a namehash implementation yet. We could instead include a new package [eth-ens-namehash](https://github.com/ensdomains/eth-ens-namehash) and use it in both places. What do you think?

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -105,7 +107,7 @@ class Web3Adapter implements EthAdapter {
     return getGnosisSafeProxyFactoryContractInstance(safeVersion, proxyFactoryContract)
   }
 
-  getContract(address: string, abi: AbiItem[]): any {
+  getContract(address: string, abi: any): any {
```
iamacook, line 110: Please avoid using `any`.

@germartinez, we should add a linting rule to prevent this.

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -125,13 +127,24 @@ class Web3Adapter implements EthAdapter {
     return this.#web3.eth.sign(message, this.#signerAddress)
   }
 
-  estimateGas(transaction: EthAdapterTransaction, options?: string): Promise<number> {
+  estimateGas(transaction: EthAdapterTransaction, options?: any): Promise<number> {
```
iamacook, line 130: Here is another instance of `any`.

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
iamacook: @usame-algan, have you checked to see if we use one in the Safe?

iamacook (commented):
Looking good so far! Just a couple type adjustments.

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
usame-algan: Good catch! As discussed we will deal with it in https://github.com/gnosis/safe-core-sdk/issues/138

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
usame-algan: done!

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
usame-algan: It looks like this is the first place where we need the namehash for the reverse lookup. Ethers can do the reverse lookup without the namehash so this only affects web3 atm.

katspaugh: @germartinez isn't ethers.js bundled with the Core SDK in any case? Or is there tree-shaking in place?

packages/safe-core-sdk/src/ethereumLibs/EthersAdapter.ts
```diff
@@ -151,6 +151,14 @@ class EthersAdapter implements EthAdapter {
   call(transaction: EthAdapterTransaction): Promise<string> {
     return this.#provider.call(transaction)
   }
+
+  ensLookup(name: string): Promise<string | null> {
+    return this.#provider.resolveName(name)
+  }
+
+  async ensReverseLookup(address: string): Promise<string | null> {
+    return await this.#provider.lookupAddress(address);
```
katspaugh, line 160: @usame-algan could you remove async-await here for consistency with the methods above?

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
katspaugh: `eth-ens-namehash` is very big too. I think it's better if use ethers temporarily. Web3 will include this function in one of the upcoming versions as per their GitHub.

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -140,7 +140,7 @@ class Web3Adapter implements EthAdapter {
   }
 
   async ensReverseLookup(address: string): Promise<string> {
-    const lookup = address.toLowerCase().substr(2) + '.addr.reverse'
+    const lookup = address.slice(-40) + '.addr.reverse'
```
katspaugh, line 143: `slice(2)` would have the same effect :)

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
germartinez: Let's not import ethers, but @ethersproject/hash sub-module instead (https://www.npmjs.com/package/@ethersproject/hash?activeTab=readme)

usame-algan: Good Idea! I removed the `eth-ens-namehash` and used the sub-module instead.

packages/safe-core-sdk/src/ethereumLibs/EthersAdapter.ts
```diff
@@ -151,6 +151,14 @@ class EthersAdapter implements EthAdapter {
   call(transaction: EthAdapterTransaction): Promise<string> {
     return this.#provider.call(transaction)
   }
+
+  ensLookup(name: string): Promise<string | null> {
```
iamacook, line 155: Are you defining the return type as `string | null` because you expect it to throw? Or does it actually return `null` when the name is not found? If it's the former, you need not specify the `null`, e.g. `Promise<string>`

This goes for `ensReverseLookup` below as well.

usame-algan: The Ethers type definitions we use specify `Promise<string | null>` as return value for both lookups so we had to change it here. The [official documentation](https://docs.ethers.io/v5/single-page/#/v5/api/providers/provider/-%23-Provider-lookupAddress) only has `Promise<string>` defined though.

iamacook: Alright, I'll resolve this, but can you add it to the issue we created the other day about updating the types?

packages/safe-core-sdk/src/ethereumLibs/EthersAdapter.ts
```diff
@@ -152,12 +153,24 @@ class EthersAdapter implements EthAdapter {
...
+      throw new Error(ErrorCodes._100)
+    }
+
+    return address
   }
 
-  ensReverseLookup(address: string): Promise<string | null> {
-    return this.#provider.lookupAddress(address);
+  async ensReverseLookup(address: string): Promise<string> {
+    const name = await this.#provider.lookupAddress(address);
```
katspaugh, line 167: This should be in a try-catch, like the web3 version. Otherwise it can throw its own error (e.g. network down).

packages/safe-core-sdk/src/ethereumLibs/EthersAdapter.ts
```diff
@@ -152,12 +153,24 @@ class EthersAdapter implements EthAdapter {
     return this.#provider.call(transaction)
   }
 
-  ensLookup(name: string): Promise<string | null> {
-    return this.#provider.resolveName(name)
+  async ensLookup(name: string): Promise<string> {
+    const address = await this.#provider.resolveName(name)
```
katspaugh, line 157: Same here.

iamacook (approved):
🚀

packages/safe-core-sdk/src/ethereumLibs/EthersAdapter.ts
```diff
@@ -151,6 +152,35 @@ class EthersAdapter implements EthAdapter {
...
+      if (!address) {
+        throw new Error(ErrorCodes._100)
+      }
+
+      return address
+    } catch (error) {
+      throw new Error(ErrorCodes._100)
+    }
+
+  }
```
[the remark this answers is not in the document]
usame-algan: I like this more than throwing twice!

katspaugh (approved):
🚀

PR #148, opened 2022-01-19
Fix Web3 tests

### LINKED ISSUE #147: Fix Web3 tests
## Context / issue
Current state of the tests does not allow to run them initializing the SDK with `web3`. Only `ethers.js` works.

`ethers.js` works because it doesn't need the ABI of a contract in order to instantiate it, but `web3` does.

After the integration of `safe-deployments`, `web3` tests started to fail because only the ABI of the contracts deployed in the networks included in `safe-deployments` are found. `safe-deployemtns`can't find the ABIs for contracts deployed in the local network (tests).

### To do:
1. Get the ABI of the Safe related contracts deployed to the local network during tests.
2. Add the ABIs to the type `ContractNetworksConfig`
3. Pass the ABIs using optional props to the methods where the contracts are instantiated.
4. Add `web3` tests to CI for Safe v1.2.0 and v1.3.0

germartinez:
## What it solves
Resolves #147

### DIFF
--- a/packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
+++ b/packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
@@ -1,5 +1,4 @@
 import { BigNumber } from '@ethersproject/bignumber'
-import { SafeVersion } from '../contracts/config'
 import {
   getGnosisSafeProxyFactoryContractInstance,
   getMultiSendContractInstance,
@@ -14,7 +13,11 @@ import {
   getSafeProxyFactoryContractDeployment
 } from '../contracts/safeDeploymentContracts'
 import { AbiItem } from '../types'
-import EthAdapter, { EthAdapterTransaction, GetSafeContractProps } from './EthAdapter'
+import EthAdapter, {
+  EthAdapterTransaction,
+  GetContractProps,
+  GetSafeContractProps
+} from './EthAdapter'
 
 export interface Web3AdapterConfig {
   /** web3 - Web3 library */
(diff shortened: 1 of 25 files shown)


packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
```diff
@@ -65,33 +69,35 @@ class Web3Adapter implements EthAdapter {
     }
     const safeContract = this.getContract(
       contractAddress,
-      safeSingletonDeployment?.abi as AbiItem[]
+      customContractAbi ?? (safeSingletonDeployment?.abi as AbiItem[])
```
iamacook, line 72: Do you have to cast this because of the conflicting types?

iamacook: It seems you have to do this a few times in this file?

iamacook (approved):
Looks good dude! I just have one question about the casting to `AbiItem[]`. I think it's related to [this](https://github.com/gnosis/safe-core-sdk/issues/138) though?

packages/safe-core-sdk/src/ethereumLibs/Web3Adapter.ts
germartinez: I did the cast because the `abi` is obtained from the `safe-deployment` library, where the object returned has an `abi` with type `any[]`:
```js
export interface SingletonDeployment {
    defaultAddress: string;
    version: string;
    abi: any[];
    networkAddresses: Record<string, string>;
    contractName: string;
    released: boolean;
}
```
To instantiate the contract I call the `getContract(address: string, abi: AbiItem[])` method, where an `AbiItem[]` is expected.

PR #158, opened 2022-01-26
Fix web3 and ethers types

germartinez:
## What it solves
Resolves https://github.com/gnosis/safe-core-sdk/issues/138

## How this PR fixes it
- Removes `any` type from `web3` and `ethers` variables
- Removes `AbiItem` local type
- Fixes tests by removing the `deterministicDeployment` property from `ERC20Mintable` contract deployment (hardhat)
- Updates deps


packages/safe-core-sdk/tests/utils/setupEthAdapter.ts
```diff
@@ -9,7 +9,7 @@ export async function getEthAdapter(signer: Signer): Promise<EthAdapter> {
   switch (process.env.ETH_LIB) {
     case 'web3':
       const signerAddress = await signer.getAddress()
-      const web3AdapterConfig: Web3AdapterConfig = { web3, signerAddress }
+      const web3AdapterConfig: Web3AdapterConfig = { web3: web3 as any, signerAddress }
```
germartinez, line 12: This is needed. `web3` instance is coming from `hardhat` package (not `web3`) and tests would complain telling that "is not assignable".

packages/safe-core-sdk-types/src/contracts/GnosisSafeContract.ts
```diff
@@ -22,6 +22,6 @@ export interface GnosisSafeContract {
     safeTransaction: SafeTransaction,
     options?: TransactionOptions
   ): Promise<TransactionResult>
-  encode(methodName: any, params: any): string
+  encode(methodName: string, params: any): string
```
iamacook, line 25: Can you type this `any`?

packages/safe-web3-lib/src/Web3Adapter.ts
```diff
@@ -99,15 +98,15 @@ class Web3Adapter implements EthAdapter {
     return getGnosisSafeProxyFactoryContractInstance(safeVersion, proxyFactoryContract)
   }
 
-  getContract(address: string, abi: AbiItem[]): any {
-    return new this.#web3.eth.Contract(abi, address)
+  getContract(address: string, abi: AbiItem | AbiItem[], options?: ContractOptions): any {
```
iamacook, line 101: Does this really return `any`? 👀

packages/safe-web3-lib/src/Web3Adapter.ts
```diff
@@ -119,7 +118,7 @@ class Web3Adapter implements EthAdapter {
     return this.#web3.eth.sign(message, this.#signerAddress)
   }
 
-  estimateGas(transaction: EthAdapterTransaction, options?: string): Promise<number> {
+  estimateGas(transaction: EthAdapterTransaction, options?: any): Promise<number> {
```
iamacook, line 121: `any` options?

packages/safe-core-sdk-types/src/contracts/GnosisSafeContract.ts
```diff
@@ -22,6 +22,6 @@ export interface GnosisSafeContract {
...
   estimateGas(methodName: string, params: any[], options: TransactionOptions): Promise<number>
```
iamacook, line 26: And these `params`?

iamacook (commented):
`any` is not your friend 🤣

packages/safe-core-sdk-types/src/contracts/GnosisSafeContract.ts
iamacook: After discussing this, I'm happy to resolve this with `any`. You may be able to strictly type it via using generics and linking `methodName` to a predefined type of strings. When `T extends <methodNameHere>` you can use a ternary to select the `params`. It's complex so I'll leave it up to you if you want to include it.

packages/safe-core-sdk-types/src/contracts/GnosisSafeContract.ts
iamacook: I'm resolving this after discussing it. Please see my comment above for the reasoning.

packages/safe-web3-lib/src/Web3Adapter.ts
iamacook: It seems like the type is `'TODO:'` on web3's end. It's possible to use `ReturnType<typeof this.#web3.eth.Contract>` to infer it I think. I'll resolve this, however, as it is not properly typed by a third party.

iamacook (approved):
Thanks for discussing the remaining issues with me. Looks great!

PR #159, opened 2022-01-28
Chore: use yarn cache in the CI script

katspaugh:
## What it solves
I've copied the same `yarn install` job as in safe-react.
It correctly uses a frozen lock file which ensures idempotent builds.


germartinez (approved):
Thank you @katspaugh! :)

PR #162, opened 2022-01-30
Fixed import-statement of 'Safe' that is 'export default' now

itinance:
Fixed README of `safe-core-sdk`-package regarding the import of `Safe` 

Wrong:
`import { Safe, SafeFactory, SafeAccountConfig } from '@gnosis.pm/safe-core-sdk'`

Correct:
`import Safe, { SafeFactory, SafeAccountConfig } from '@gnosis.pm/safe-core-sdk'` 

because Safe is `export default`: https://github.com/gnosis/safe-core-sdk/blob/main/packages/safe-core-sdk/src/Safe.ts#L602

This PR fixes the corresponding example in the README.


itinance:

germartinez (approved):
Good catch! Thank you for fixing it :)

PR #166, opened 2022-02-02
Fix safe version and refactor safe deployments functions

### LINKED ISSUE #165: SafeFactory only deploys version 1.3.0
## Description

I'd like to deploy older safes to test my application when users are using older safe contract versions. However, I cannot deploy a safe lower than version 1.3.0. 

## Environment

- Safe Core SDK version: 2.0.0
- Safe contract version: 1.1.1 (desired)
- Environment:
  - browser: chrome

## Steps to reproduce

```
const safeVersion = '1.1.1'
const safeFactory = SafeFactory.create({
    ethAdapter,
    safeVersion,
  });
const safeAccountConfig: SafeAccountConfig = {
    owners: ['0xa...', '0xb...'],
    threshold: 1,
};
const safe = await safeFactory.deploySafe({ safeAccountConfig });
console.log(await safe.getContractVersion())
```
The resulting safe created is always contract version 1.3.0

## Expected result

I expect the contract version to match the `safeVersion` input to `SafeFactory.create`

germartinez:
## What it solves
Resolves #165 

## How this PR fixes it:
- In the `SafeFactory` class, inside the method `getContract` was a call to the method `getSafeContractDeployment` where the safeVersion was set to the last current version. This was wrong.

- Refactor the methods that return the instances of the Safe contracts:
  - The methods `getProxyFactoryContract`, `getMultiSendContract` and `getSafeContract` were extracted to a different file.
  - There were two different implementations of `getSafeContract` before (one with `customSafeAddress` as a prop and other without it). These two were also merged in one.

### DIFF
--- a/packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
+++ b/packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
@@ -86,7 +86,7 @@ export async function getSafeContract({
   })
   const isContractDeployed = await ethAdapter.isContractDeployed(gnosisSafeContract.getAddress())
   if (!isContractDeployed) {
-    throw new Error('Safe Proxy contract is not deployed in the current network')
+    throw new Error('Safe Proxy contract is not deployed on the current network')
   }
   return gnosisSafeContract
 }
@@ -109,7 +109,7 @@ export async function getProxyFactoryContract({
     safeProxyFactoryContract.getAddress()
   )
   if (!isContractDeployed) {
-    throw new Error('Safe Proxy Factory contract is not deployed in the current network')
+    throw new Error('Safe Proxy Factory contract is not deployed on the current network')
   }
   return safeProxyFactoryContract
 }
@@ -130,7 +130,7 @@ export async function getMultiSendContract({
   })
   const isContractDeployed = await ethAdapter.isContractDeployed(multiSendContract.getAddress())
   if (!isContractDeployed) {
-    throw new Error('Multi Send contract is not deployed in the current network')
+    throw new Error('Multi Send contract is not deployed on the current network')
   }
   return multiSendContract
 }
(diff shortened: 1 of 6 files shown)


packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
```diff
@@ -7,8 +13,32 @@ import {
   getSafeSingletonDeployment,
   SingletonDeployment
 } from '@gnosis.pm/safe-deployments'
+import { ContractNetworkConfig } from '../types'
 import { safeDeploymentsL1ChainIds, safeDeploymentsVersions } from './config'
 
+interface GetSafeContractInstanceProps {
```
iamacook, line 19: There seems to be some repetition in these types? Maybe it's worthwhile making a base `ContractInstanceProps` type?

packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
```diff
@@ -37,3 +67,70 @@ export function getSafeProxyFactoryContractDeployment(
...
+  const gnosisSafeContract = ethAdapter.getSafeContract({
+    safeVersion,
+    chainId,
+    singletonDeployment,
+    customContractAddress: customSafeAddress ?? customContracts?.safeMasterCopyAddress,
+    customContractAbi: customContracts?.safeMasterCopyAbi
+  })
+  const isContractDeployed = await ethAdapter.isContractDeployed(gnosisSafeContract.getAddress())
+  if (!isContractDeployed) {
+    throw new Error('Safe Proxy contract is not deployed in the current network')
```
iamacook, line 89: The three error messages here should be "...deployed on the current network"

iamacook (approved):
Looks good to me 🚀  Just a couple of suggestions

iamacook (approved):
💪

PR #167, opened 2022-02-02
Development

germartinez:
Fixes #166

### DIFF
--- a/packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
+++ b/packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
@@ -1,4 +1,10 @@
-import { SafeVersion } from '@gnosis.pm/safe-core-sdk-types'
+import {
+  EthAdapter,
+  GnosisSafeContract,
+  GnosisSafeProxyFactoryContract,
+  MultiSendContract,
+  SafeVersion
+} from '@gnosis.pm/safe-core-sdk-types'
 import {
   DeploymentFilter,
   getMultiSendDeployment,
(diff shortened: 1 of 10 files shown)


packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
```diff
@@ -37,3 +56,70 @@ export function getSafeProxyFactoryContractDeployment(
...
+}: GetSafeContractInstanceProps): Promise<GnosisSafeContract> {
+  const singletonDeployment = getSafeContractDeployment(safeVersion, chainId, isL1SafeMasterCopy)
+  const gnosisSafeContract = ethAdapter.getSafeContract({
+    safeVersion,
+    chainId,
+    singletonDeployment,
+    customContractAddress: customSafeAddress ?? customContracts?.safeMasterCopyAddress,
+    customContractAbi: customContracts?.safeMasterCopyAbi
+  })
+  const isContractDeployed = await ethAdapter.isContractDeployed(gnosisSafeContract.getAddress())
```
iamacook, line 76: I feel like you could abstract this error-handling logic as it's repeated, but I'm not entirely sure on the best approach. Food for thought.

iamacook (approved):

packages/safe-core-sdk/src/contracts/safeDeploymentContracts.ts
germartinez: I thought the same at the beginning, but I didn't see a real benefit in extracting the call to `isContractDeployed` and the following if statement to a new function given that the error message would also be a parameter of this small new function because it is different depending on the contract.
Other options could be moving the exception to `isContractDeployed` but it has no sense, it returns a boolean.
So I left it as it it now.
Any better idea?

iamacook: It's not repeated so often and it's clean enough. I think it's fine to leave it as is.

PR #178, opened 2022-02-22
EIP-3770

### LINKED ISSUE #74: Allow EIP-3770 chain-specific addresses
## Overview

We are pushing to [implement EIP-3770](https://github.com/gnosis/safe/issues/487) address references, which include a prefixed chain id before an address. We need to allow users to input addresses with a chain prefix.

image

## Goals
- Accept/validate EIP-3770 address input.

## Requirements
- Modify inputs/validation to allow EIP-3770 address formats.
- Validate networks against this [map](https://chainid.network/shortNameMapping.json). (A service will likely be created for this but it is acceptable to do this client-side in the web app.)
- Validate that prefixed addresses match those of the current network.

## Links
[Kick off notes](https://docs.google.com/document/d/1iUuNO8ObL45A6lTper1HdZPtZyY88qm_B27g2v2z4gU/edit)

germartinez:
## What it solves
Resolves #74 

- Implements EIP-3770 in `safe-core-sdk-utils`


packages/safe-core-sdk-utils/src/eip-3770/config.ts
```diff
@@ -0,0 +1,38 @@
+interface NetworkShortName {
+  shortName: string
+  chainId: number
+}
+
+export const networks: NetworkShortName[] = [
```
iamacook, line 6: Are these related to contracts we currently have deployed?

iamacook (commented):
Nice! Very thorough testing

iamacook (approved):

packages/safe-core-sdk-utils/src/eip-3770/config.ts
germartinez: Yes, I took them from `safe-deployments`

PR #180, opened 2022-02-22
Update github actions event triggers

germartinez:
## What it solves
Updates the event triggers in the Github actions to reduce the number of builds and duplicate tests

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,13 +1,10 @@
 name: e2e Test
 on:
+  pull_request:
   push:
     branches:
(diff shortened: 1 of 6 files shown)


iamacook (approved):
Looks good. I assume:

```yml
this:
falls through:
  - to here
```

germartinez:
@iamacook yes, checking all the github actions that were executed in this PR (see below), it is working under the following condition:

When the event `pull_request` (on any branch) or `push` (only to `development` or `main` branches) are triggered, the github actions will execute.

Before this PR, actions were always executing twice.

PR #184, opened 2022-03-08
Sync EIP-3770 networks with safe-deployment networks

germartinez:
## What it solves
In different packages inside the monorepo:
- Updates `safe-deployments` dependency to v1.10.0

In `safe-core-sdk-utils`:
- Adds a script that is executed when the project is built and checks if there are networks from `safe-deployments` that are not included in the local EIP-3770 configuration
- Adds missing networks to `networks` config


packages/safe-core-sdk-utils/scripts/checkSafeDeployments.ts
```diff
@@ -0,0 +1,29 @@
+import { DeploymentFilter, getSafeSingletonDeployment } from '@gnosis.pm/safe-deployments'
+import { networks } from '../src/eip-3770/config'
+
+const safeVersion = '1.3.0'
```
iamacook, line 4: Do you have an enum for Safe versions in the SDK?

iamacook (approved):

usame-algan (approved):
👍

francovenica:
Is there is something to check here in the web app? If there is nothing specific we can move it to QA done.

PR #187, opened 2022-03-30
Update deps

germartinez:
## What it solves
- Update old deps
- Adds new networks from `safe-deployments` (EIP-3770 configuration)


usame-algan (approved):
LGTM!

PR #188, opened 2022-04-04
Add a callback to Safe deployment

germartinez:
## What it solves
Adds a callback to the `deploySafe` method in the `SafeFactory` class.

This callback receives as a parameter the `tcxHash` of the deployment transaction, allowing to track the transaction outside of the SDK, in parallel, meanwhile the deployment is being executed.

```js
const callback = (txHash: string): void => {
  console.log({ txHash })
}

const safeFactory = await SafeFactory.create({ ethAdapter })

const deploySafeProps: DeploySafeProps = { safeAccountConfig, callback }
const safe = await safeFactory.deploySafe(deploySafeProps)
```


iamacook (approved):
Looks good to me!

usame-algan (approved):
Very nice 💯

PR #208, opened 2022-06-15
Support Sign Typed Data

### LINKED ISSUE #108: Support EIP-712 signatures
## Context / issue

User should understand what they sign therefore we should support typed data signing

germartinez:
## What it solves
Resolves #108 

## How this PR fixes it
Safe Core SDK has the new method `signTypedData` that works according to the [EIP-712](https://eips.ethereum.org/EIPS/eip-712):
```js
async signTypedData(safeTransaction: SafeTransaction, methodVersion?: 'v3' | 'v4'): Promise<SafeSignature>
```

The method `signTransaction` was updated to allow chosing the method to sign:
```js
async signTransaction(
  safeTransaction: SafeTransaction,
  signingMethod: 'eth_sign' | 'eth_signTypedData' = 'eth_sign'
): Promise<void>
```


packages/safe-core-sdk-utils/src/eip-712/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export function generateTypedData({
+  safeAddress,
+  safeVersion,
+  chainId,
+  safeTransactionData
+}: SafeTransactionEIP712Args): GenerateTypedData {
+  const eip712WithChainId = semverSatisfies(safeVersion, '>=1.3.0')
```
iamacook, line 52: I'd suggest extracting this to a constant. It's used in the function above as well.

packages/safe-ethers-lib/src/EthersAdapter.ts
```diff
@@ -137,6 +138,16 @@ class EthersAdapter implements EthAdapter {
     return this.#signer.signMessage(messageArray)
   }
 
+  async signTypedData(safeTransactionEIP712Args: SafeTransactionEIP712Args): Promise<string> {
+    const typedData = generateTypedData(safeTransactionEIP712Args)
+    const signature = await (this.#signer as JsonRpcSigner)._signTypedData(
```
iamacook, line 143: Instead of casting, can't you initially type `#signer` as `JsonRpcSigner` instead of signer?

packages/safe-web3-lib/src/Web3Adapter.ts
```diff
@@ -134,6 +135,44 @@ class Web3Adapter implements EthAdapter {
...
+      params:
+        methodVersion === 'v3' || methodVersion === 'v4'
+          ? [this.#signerAddress, jsonTypedData]
+          : [jsonTypedData, this.#signerAddress],
+      from: this.#signerAddress,
+      id: new Date().getTime()
+    }
+    return new Promise((resolve, reject) => {
+      const provider = this.#web3.currentProvider as AbstractProvider
+      provider.sendAsync(signedTypedData, (err, signature) => {
```
iamacook, line 162: Don't all providers have the `sendAsync` method?

packages/safe-core-sdk-utils/src/eip-712/index.ts
germartinez: I don't see how a constant could apply here (`safeVersion` is a variable). Maybe a function but I don't think that is worthy

germartinez: Oh! you probably mean ">=1.3.0" :D

packages/safe-ethers-lib/src/EthersAdapter.ts
iamacook: Maybe you chould add a type guard to check that it supports the `_signTypedData` method?

iamacook (approved):
🚀

PR #216, opened 2022-07-11
Support endpoint to get all the Safes with a given module enabled

### LINKED ISSUE #207: Support new endpoint in Transaction Service for Safe modules
## Context / issue

There is a new endpoint in the Safe Transaction Service that returns a list of the Safes that have enabled a specific module.

Check it here: https://github.com/safe-global/safe-transaction-service/issues/893

## Proposed solution

Add support for this endpoint in the Safe Service Client package.

germartinez:
## What it solves
Resolves #207


usame-algan (approved):
Love the tests!

PR #220, opened 2022-07-14
Fix signTypedData

germartinez:
## What it solves

Fixes the error `[Error: invalid BigNumber value (argument="value", value=undefined, code=INVALID_ARGUMENT, version=bignumber/5.6.2)](https://ethereum.stackexchange.com/questions/130776/error-invalid-bignumber-value-argument-value-value-undefined-code-invalid)` for Safes with version `<v1.3.0`.

This issue was caused because `chainId` had a wrong type making `signer._signTypedData()` complain about it.

### DIFF
--- a/packages/safe-core-sdk-utils/src/eip-712/index.ts
+++ b/packages/safe-core-sdk-utils/src/eip-712/index.ts
@@ -55,7 +55,6 @@ export function generateTypedData({
   const typedData: GenerateTypedData = {
     types: getEip712MessageTypes(safeVersion),
     domain: {
-      chainId: eip712WithChainId ? chainId : undefined,
       verifyingContract: safeAddress
     },
     primaryType: 'SafeTx',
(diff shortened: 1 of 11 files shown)


packages/safe-core-sdk-utils/src/eip-712/index.ts
```diff
@@ -68,5 +67,8 @@ export function generateTypedData({
       nonce: BigNumber.from(safeTransactionData.nonce)
     }
   }
+  if (eip712WithChainId) {
+    typedData.domain.chainId = chainId
```
schmanu, line 71: I guess the error was caused because ethers would try to encode `undefined` instead of ignoring it if the key is present?

Would it be possible to unit-test this? I feel like this is missed / refactored wrong easily.

germartinez: Yes, that was exactly the issue ;)
I'll add some tests to cover that

usame-algan (approved):
LGTM 👍

PR #229, opened 2022-07-28
Calculate counterfactual Safe

### LINKED ISSUE #215: Add method to predict the deployment address of a Safe
## Context / issue

`GnosisSafeProxyFactoryEthersContract` and `GnosisSafeProxyFactoryWeb3Contract` classes from `safe-ethers-lib` and `safe-web3-lib` packages respectively implement the method `createProxy` that deploys a new Safe proxy account, but there is currently no implementation to predict the address where a Safe will be deployed.

## Proposed solution

Add a new method called `predictSafeAddress`:

```js
itnerface SafeAccountConfig {
  owners: string[]
  threshold: number
  to?: string
  data?: string
  fallbackHandler?: string
  paymentToken?: string
  payment?: number
  paymentReceiver?: string
}

interface SafeDeploymentConfig {
  saltNonce: number
}

class SafeFactory {
  ...

  predictSafeAddress(
    safeAccountConfig: SafeAccountConfig,
    safeDeploymentConfig: SafeDeploymentConfig
  ): string {
    // TO-DO
    return predictedSafeAddress
  }
}
```

germartinez:
## What it solves
Resolves #215

### DIFF
--- a/packages/safe-core-sdk/package.json
+++ b/packages/safe-core-sdk/package.json
@@ -48,6 +48,7 @@
     "@nomiclabs/hardhat-web3": "^2.0.0",
     "@types/chai": "^4.3.0",
     "@types/chai-as-promised": "^7.1.5",
+    "@types/ethereumjs-abi": "^0.6.3",
     "@types/mocha": "^9.1.0",
     "@types/node": "^17.0.23",
     "@types/semver": "^7.3.9",
@@ -91,6 +92,7 @@
     "@ethersproject/solidity": "^5.6.0",
     "@gnosis.pm/safe-core-sdk-types": "^1.2.1",
     "@gnosis.pm/safe-deployments": "1.15.0",
+    "ethereumjs-abi": "^0.6.8",
     "ethereumjs-util": "^7.1.4",
     "semver": "^7.3.5",
     "web3-utils": "^1.7.1"
--- a/packages/safe-core-sdk/package.json
+++ b/packages/safe-core-sdk/package.json
@@ -48,7 +48,6 @@
     "@nomiclabs/hardhat-web3": "^2.0.0",
     "@types/chai": "^4.3.0",
     "@types/chai-as-promised": "^7.1.5",
-    "@types/ethereumjs-abi": "^0.6.3",
     "@types/mocha": "^9.1.0",
     "@types/node": "^17.0.23",
     "@types/semver": "^7.3.9",
@@ -92,7 +91,6 @@
     "@ethersproject/solidity": "^5.6.0",
     "@gnosis.pm/safe-core-sdk-types": "^1.2.1",
     "@gnosis.pm/safe-deployments": "1.15.0",
-    "ethereumjs-abi": "^0.6.8",
     "ethereumjs-util": "^7.1.4",
     "semver": "^7.3.5",
     "web3-utils": "^1.7.1"
--- a/packages/safe-core-sdk/src/safeFactory/index.ts
+++ b/packages/safe-core-sdk/src/safeFactory/index.ts
@@ -157,7 +157,7 @@ class SafeFactory {
 
     const initializer = await this.encodeSetupCallData(safeAccountConfig)
     const saltNonce = safeDeploymentConfig.saltNonce
-    const encodedNonce = toBuffer(this.#ethAdapter.encodeParameter('uint256', saltNonce)).toString(
+    const encodedNonce = toBuffer(this.#ethAdapter.encodeParameters(['uint256'], [saltNonce])).toString(
       'hex'
     )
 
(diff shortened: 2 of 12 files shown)


packages/safe-core-sdk/src/safeFactory/index.ts
```diff
@@ -140,13 +147,45 @@ class SafeFactory {
...
+  }: PredictSafeProps): Promise<string> {
+    validateSafeAccountConfig(safeAccountConfig)
+    if (safeDeploymentConfig) {
+      validateSafeDeploymentConfig(safeDeploymentConfig)
+    }
+
+    const from = this.#safeProxyFactoryContract.getAddress()
+
+    const initializer = await this.encodeSetupCallData(safeAccountConfig)
+    const saltNonce = safeDeploymentConfig.saltNonce.toString()
```
mikhailxyz, line 162: can you make it a string in the config? it's a `uint256` and javascript number cannot fit it

packages/safe-core-sdk/tests/safeFactory.test.ts
```diff
@@ -95,6 +96,93 @@ describe('Safe Proxy Factory', () => {
     })
   })
 
+  describe('predictSafeAddress', async () => {
+    it('should fail if there are no owners', async () => {
```
mikhailxyz, line 100: I think it is possible to deploy a safe with no owners and initialize them later so not sure if we want to fail in this case

mikhailxyz: https://github.com/safe-global/safe-contracts/blob/main/contracts/proxies/GnosisSafeProxyFactory.sol#L67

packages/safe-core-sdk/tests/safeFactory.test.ts
```diff
@@ -95,6 +96,93 @@ describe('Safe Proxy Factory', () => {
...
+        contractNetworks
+      })
+      const owners = [account1.address, account2.address]
+      const threshold = 2
+      const safeAccountConfig: SafeAccountConfig = { owners, threshold }
+      const safeDeploymentConfig: SafeDeploymentConfig = { saltNonce: -1 }
+      const predictSafeProps: PredictSafeProps = { safeAccountConfig, safeDeploymentConfig }
+      chai
+        .expect(safeFactory.predictSafeAddress(predictSafeProps))
+        .rejectedWith('saltNonce must be greater than 0')
```
mikhailxyz, line 161: cant it equal 0?

packages/safe-core-sdk/src/safeFactory/index.ts
```diff
@@ -5,12 +5,14 @@ import {
   SafeVersion,
   TransactionOptions
 } from '@gnosis.pm/safe-core-sdk-types'
+import abi from 'ethereumjs-abi'
```
mikhailxyz, line 8: Both `ethers` and `web3` have utilities for abi encoding. Can we add corresponding functions to the adapters? Since the library can be used in-browser, we should be cautious about blowing the bundle size imo

packages/safe-core-sdk/package.json
```diff
@@ -48,6 +48,7 @@
     "@nomiclabs/hardhat-web3": "^2.0.0",
     "@types/chai": "^4.3.0",
     "@types/chai-as-promised": "^7.1.5",
+    "@types/ethereumjs-abi": "^0.6.3",
```
mikhailxyz, line 51: I think you don't need it anymore. please check

packages/safe-web3-lib/src/Web3Adapter.ts
```diff
@@ -187,6 +191,10 @@ class Web3Adapter implements EthAdapter {
   call(transaction: EthAdapterTransaction): Promise<string> {
     return this.#web3.eth.call(transaction)
   }
+
+  encodeParameter(type: string, value: any): string {
```
mikhailxyz, line 195: What do you think about using `encodeParameters` instead of `encodeParameter` and accepting an array of types and values? If we proceed with this function if any time in the future, we need to implement the same function, but for encoding multiple parameters, we would either have to do a breaking change or add a new function

mikhailxyz: https://web3js.readthedocs.io/en/v1.2.11/web3-eth-abi.html#encodeparameters

packages/safe-web3-lib/src/contracts/GnosisSafeProxyFactory/GnosisSafeProxyFactoryWeb3Contract.ts
```diff
@@ -20,16 +20,20 @@ class GnosisSafeProxyFactoryWeb3Contract implements GnosisSafeProxyFactoryContra
...
     safeMasterCopyAddress,
     initializer,
     saltNonce,
     options,
     callback
   }: CreateProxyProps): Promise<string> {
-    if (saltNonce < 0) {
-      throw new Error('saltNonce must be greater than 0')
-    }
+    //if (saltNonce < 0) {
```
mikhailxyz, line 34: why comment?

mikhailxyz (approved):
good job lad

PR #233, opened 2022-08-01
Make signTransaction method immutable

### LINKED ISSUE #228: Make signTransaction method immutable
## Context / issue

`signTransaction()` method modifies the param that is received.

## Proposed solution

It's a better practice that signing a transaction returns a new transaction object without mutating the passed argument.

germartinez:
## What it solves
Resolves #228


packages/safe-core-sdk/README.md
```diff
@@ -108,7 +108,7 @@ Before executing this transaction, it must be signed by the owners and this can
 The `owner1` account signs the transaction off-chain.
 
 ```js
-const owner1Signature = await safeSdk.signTransaction(safeTransaction)
+const signedSafeTransaction = await safeSdk.signTransaction(safeTransaction)
 ```
 
 Because the signature is off-chain, there is no interaction with the contract and the signature becomes available at `safeTransaction.signatures`.
```
usame-algan, line 114: Should we rephrase this as well to include `signedSafeTransaction`?

germartinez: True. Good catch, thanks!

usame-algan (approved):

PR #234, opened 2022-08-01
Support EIP-1559

### LINKED ISSUE #206: Support EIP-1559
## Context / issue

[EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) is not currently supported

## Proposed solution

Add support for the [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md)

germartinez:
## What it solves
Resolves #206

### DIFF
--- a/packages/safe-core-sdk-types/src/ethereumLibs/EthAdapter.ts
+++ b/packages/safe-core-sdk-types/src/ethereumLibs/EthAdapter.ts
@@ -13,6 +13,8 @@ export interface EthAdapterTransaction {
   value?: string
   gasPrice?: number
   gasLimit?: number
+  maxFeePerGas?: number | string
+  maxPriorityFeePerGas?: number | string
 }
 
 export interface GetContractProps {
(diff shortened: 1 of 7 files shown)


packages/safe-core-sdk-types/src/ethereumLibs/EthAdapter.ts
```diff
@@ -13,6 +13,8 @@ export interface EthAdapterTransaction {
   value?: string
   gasPrice?: number
   gasLimit?: number
+  maxFeePerGas?: number | string
+  maxPriorityFeePerGas?: number | string
```
usame-algan, line 17: Is there a reason we support `string` here but not for `gasPrice`?

germartinez: Given the regular values that `gasPrice` takes, `number` is enough. However, it should be a `string | number` like the properties that were added in this PR.
I will recheck all the types of the similar properties and update them in a new PR

usame-algan (approved):
👍

PR #232, opened 2022-08-01
Migrate Rinkeby tests to Goerli

germartinez:
## What it solves
Resolves #231

PR #235, opened 2022-08-01
v2.4.0

germartinez:
## What it solves
- #232 
- #234 
- #233


usame-algan (approved):

packages/safe-core-sdk/scripts/postinstall.ts
```diff
@@ -0,0 +1,7 @@
+function postInstall() {
+  console.warn(
+    'WARNING! "signTransaction" method now *returns* the signed Safe transaction. Update your code according to the new documentation: https://github.com/safe-global/safe-core-sdk/tree/main/packages/safe-core-sdk#signtransaction. In >=v3.0.0 the signature will only be added to the returned object, not the one that is passed as params.'
```
katspaugh, line 3:
```suggestion
    `⚠️ the "signTransaction" method now returns a signed Safe transaction. Please update your code according to the new documentation: https://github.com/safe-global/safe-core-sdk/tree/main/packages/safe-core-sdk#signtransaction. In >=v3.0.0, the signature will be added only to the return transaction object, and not to the one passed as an argument.`

katspaugh:
@germartinez bear in mind 2.3.0 -> 2.4.0 will require people to manually update the package, so they likely won't see the warning in a while. I'd suggest to release 2.2.1 with the mutation fix and warning, and then either 3.0.0 or 2.3.0 with the rest.

packages/safe-core-sdk/scripts/postinstall.ts
germartinez: Added. Thanks @katspaugh

germartinez:
> I'd suggest to release 2.2.2 with the mutation fix and warning, and then either 3.0.0 or 2.3.0 with the rest.

Currently is v2.3.0, do you mean release 2.3.1? And later either 3.0.0 or 2.4.0 with the rest?

germartinez:
Or maybe v2.2.2 and v2.3.1?

katspaugh:
I meant 2.2.2 because 2.3.0 itself was released only 4 days ago AFAIU.

PR #244, opened 2022-08-24
Add user nonce to TransactionOptions type

germartinez:
## What it solves
Allows to specify the nonce from the sender wallet when executing a Safe transaction via the Core SDK.

### DIFF
--- a/packages/safe-web3-lib/src/Web3Adapter.ts
+++ b/packages/safe-web3-lib/src/Web3Adapter.ts
@@ -53,6 +53,10 @@ class Web3Adapter implements EthAdapter {
     return BigNumber.from(await this.#web3.eth.getBalance(address))
   }
 
+  async getNonce(address: string): Promise<number> {
+    return this.#web3.eth.getTransactionCount(address)
+  }
+
   async getChainId(): Promise<number> {
     return this.#web3.eth.getChainId()
   }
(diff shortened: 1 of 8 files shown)


packages/safe-web3-lib/src/Web3Adapter.ts
```diff
@@ -53,6 +53,10 @@ class Web3Adapter implements EthAdapter {
     return BigNumber.from(await this.#web3.eth.getBalance(address))
   }
 
+  async getNonce(address: string): Promise<number> {
```
katspaugh, line 56: Why do we need to proxy this function? If we really do, please also add a second optional param called `blockTag`.

usame-algan: What can the `blockTag` be used for?

katspaugh: In the app, we're passing `pending` to fetch unfinished txen, because that's how you know the latest nonce.

katspaugh: Actually, we should just probably pass `pending` here, like in safe-react, w/o making it a parameter.

germartinez: `getNonce` is not used directly when using the user `nonce` from the `TransactionOptions` type.
It is used in tests and it could also be useful for people playing around with the `nonce` property.

I added an optional argument that allows to specify the `blockTag`, adding it to methods that were not receiving it as well.

usame-algan (approved):
🚀

PR #245, opened 2022-08-24
Make executeTransaction immutable

germartinez:
## What it solves
Avoids modifying the transaction object that is passed to the `executeTransaction` method in the SDK.


usame-algan (approved):
👍

PR #246, opened 2022-08-24
Make signTransaction method immutable

germartinez:
## What it solves
This PR makes immutable the `signTransaction` method from the Core SDK by not adding the signature to the passed `safeTransaction` argument.


katspaugh (approved):
Thanks!

PR #249, opened 2022-08-29
fix readme of packages/safe-web3-lib 

kyosyun:
## What it solves
Only fix readme of `packages/safe-web3-lib`.
Before fix, errors are happen when use ethAdapter

```
# example
const id = await ethAdapter.getChainId()
-> TypeError: Cannot read properties of undefined (reading 'getChainId')
```


kyosyun:

germartinez (approved):
Thank you for the catch! :)

PR #250, opened 2022-08-30
Support MultiSendCallOnly

germartinez:
## What it solves
- Update config files to:
  - Generate MultiSendCallOnly typechain files.
  - Deploy MultiSendCallOnly contract with hardhat for testing.
- Update `createTransaction` method signature and logic to optionally use MultiSendCallOnly contract.
  Logic includes:
  - Handle MultiSendCallOnly contract instances.
  - Add `multisendCallOnly` in the ContractManager class.
- Update docs accordingly.
- Update tests accordingly.

### DIFF
--- a/packages/safe-core-sdk-types/src/contracts/MultiSendCallOnlyContract.ts
+++ b/packages/safe-core-sdk-types/src/contracts/MultiSendCallOnlyContract.ts
@@ -1,0 +1,4 @@
+export interface MultiSendCallOnlyContract {
+  getAddress(): string
+  encode(methodName: any, params: any): string
+}
--- a/packages/safe-core-sdk/README.md
+++ b/packages/safe-core-sdk/README.md
@@ -171,9 +171,11 @@ const safeFactory = await SafeFactory.create({ ethAdapter })
   const contractNetworks: ContractNetworksConfig = {
     [id]: {
       multiSendAddress: '<MULTI_SEND_ADDRESS>',
+      multiSendCallOnlyAddress: '<MULTI_SEND_CALL_ONLY_ADDRESS>',
       safeMasterCopyAddress: '<MASTER_COPY_ADDRESS>',
       safeProxyFactoryAddress: '<PROXY_FACTORY_ADDRESS>',
       multiSendAbi: '<MULTI_SEND_ABI>', // Optional. Only needed with web3.js
+      multiSendCallOnlyAbi: '<MULTI_SEND_CALL_ONLY_ABI>', // Optional. Only needed with web3.js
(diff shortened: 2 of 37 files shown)


packages/safe-core-sdk-types/src/contracts/MultiSendCallOnlyContract.ts
```diff
@@ -0,0 +1,4 @@
+export interface MultiSendCallOnlyContract {
+  getAddress(): string
+  encode(methodName: any, params: any): string
```
iamacook, line 3: Is using `any` necessary?

packages/safe-core-sdk/README.md
```diff
@@ -496,7 +505,14 @@ Returns a Safe transaction ready to be signed by the owners and executed. The Sa
     refundReceiver, // Optional
     nonce // Optional
   }
-  const safeTransaction = await safeSdk.createTransaction(transactions, options)
+  const safeTransaction = await safeSdk.createTransaction({ safeTransactionData, options })
+  ```
+
+  In addition, the optional `callsOnly` parameter, that is false by default, allows to force the use of the `MultiSendCallOnly` instead of the `MultiSend` contract when sending a batch transaction:
```
iamacook, line 511:
```suggestion
  In addition, the optional `callsOnly` parameter, that is `false` by default, allows to force the use of the `MultiSendCallOnly` instead of the `MultiSend` contract when sending a batch transaction:
```

packages/safe-core-sdk-types/src/contracts/MultiSendCallOnlyContract.ts
germartinez: Otherwise it gets too complex. It is one of those cases where the type of one parameter depends on the value of another parameter. There are a few similar cases we discussed in the past

packages/safe-core-sdk/README.md
```diff
@@ -496,7 +505,14 @@ Returns a Safe transaction ready to be signed by the owners and executed. The Sa
     refundReceiver, // Optional
     nonce // Optional
   }
-  const safeTransaction = await safeSdk.createTransaction(transactions, options)
+  const safeTransaction = await safeSdk.createTransaction({ safeTransactionData, options })
+  ```
+
+  In addition, the optional `callsOnly` parameter, that is `false` by default, allows to force the use of the `MultiSendCallOnly` instead of the `MultiSend` contract when sending a batch transaction:
```
katspaugh, line 511: > that is `false` by default

`which is false by default` would sound slightly better.

PR #251, opened 2022-08-30
Rename methods and error messages

germartinez:
## What it solves
Taking advantage of the next mayor release, this PR:
- Renames methods that return a Safe transaction that modifies the Safe settings.
  Example: `const safeTx = getAddOwnerTx(...)` -> `const safeTx = createAddOwnerTx(...)`
- Renames contract names in error messages
  Example: `Error: [...] Multi Send contract [...]` -> `Error: [...] MultiSend contract [...]`
- Updates tests and documentation accordingly

### DIFF
--- a/packages/safe-core-sdk/README.md
+++ b/packages/safe-core-sdk/README.md
@@ -352,7 +352,7 @@ const safeSdk2 = await safeSdk.connect({ ethAdapter, safeAddress })
 
 ### getAddress
 
-Returns the address of the current Safe Proxy contract.
+Returns the address of the current SafeProxy contract.
 
 ```js
 const address = safeSdk.getAddress()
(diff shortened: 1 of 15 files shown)


packages/safe-core-sdk/README.md
```diff
@@ -651,12 +651,12 @@ const txHash = await safeSdk.getTransactionHash(safeTransaction)
 const owners = await safeSdk.getOwnersWhoApprovedTx(txHash)
 ```
 
-### getEnableModuleTx
+### createEnableModuleTx
```
iamacook, line 654: Should we format these, i.e.

```suggestion
### `createEnableModuleTx`
```

germartinez: I'm personally not a fan of using `` in titles like these from the visuals perspective. I found ENS is doing it [this way](https://github.com/ensdomains/docs/blob/master/contract-api-reference/publicresolver.md), but also not sure which one is better. I think I prefer to keep the method names as they are, all together.

iamacook: Sure, let's keep them as is.

iamacook (approved):

PR #279, opened 2022-10-25
Check if a transaction is valid

germartinez:
## What it solves
This PR adds the method `isValidTransaction` that returns a boolean value and allows to check if a transaction will be executed with or without errors.

### DIFF
--- a/packages/safe-core-sdk/src/Safe.ts
+++ b/packages/safe-core-sdk/src/Safe.ts
@@ -367,7 +367,7 @@ class Safe {
     safeTransaction.signatures.forEach((signature) => {
       signedSafeTransaction.addSignature(signature)
     })
-    return safeTransaction
+    return signedSafeTransaction
   }
 
   /**
--- a/packages/safe-core-sdk/src/Safe.ts
+++ b/packages/safe-core-sdk/src/Safe.ts
@@ -677,6 +677,45 @@ class Safe {
   }
 
   /**
+   * Checks if a Safe transaction can be executed successfully with no errors.
+   *
+   * @param safeTransaction - The Safe transaction to check
+   * @param options - The Safe transaction execution options. Optional
+   * @returns TRUE if the Safe transaction can be executed successfully with no errors
+   */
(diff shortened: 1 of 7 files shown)


packages/safe-core-sdk/src/Safe.ts
```diff
@@ -676,6 +676,45 @@ class Safe {
...
+   *
+   * @param safeTransaction - The Safe transaction to check
+   * @param options - The Safe transaction execution options. Optional
+   * @returns TRUE if the Safe transaction can be executed successfully with no errors
+   */
+  async isValidTransaction(
+    safeTransaction: SafeTransaction,
+    options?: TransactionOptions
+  ): Promise<boolean> {
+    const signedSafeTransaction = await this.createTransaction({
```
schmanu, line 690: The first part of this function which creates a copy of the transaction could be moved to some util function.

packages/safe-core-sdk/tests/execution.test.ts
```diff
@@ -34,6 +34,72 @@ describe('Transactions execution', () => {
...
+      const tx = await safeSdk1.createTransaction({ safeTransactionData })
+      const rejectTx = await safeSdk1.createRejectionTransaction(tx.data.nonce)
+      const signedRejectTx = await safeSdk1.signTransaction(rejectTx)
+      const txRejectResponse = await safeSdk2.executeTransaction(signedRejectTx)
+      await waitSafeTxReceipt(txRejectResponse)
+      const signedTx = await safeSdk1.signTransaction(tx)
+      const isTxExecutable = await safeSdk2.isValidTransaction(signedTx)
+      await chai.expect(isTxExecutable).to.be.eq(false)
+      const safeFinalBalance = await safeSdk1.getBalance()
+      chai.expect(safeInitialBalance.toString()).to.be.eq(safeFinalBalance.toString())
```
schmanu, line 72: Nice test cases! 👍

schmanu (approved):
Cool stuff! I just have a minor suggestion but all in all this looks really good!

PR #286, opened 2022-11-11
Add networks from safe-deployments v1.17.0

germartinez:
## What it solves
- [x] Add the networks included in `safe-deployments v1.17.0` (not released yet)
- [x] update `safe-deployments` dependency once it is released on monday 14th Nov

Meanwhile, it is expected that the build fails


iamacook (approved):
Maybe it's worthwhile creating a script for this? We could parse [this list](https://github.com/ethereum-lists/chains/tree/master/_data/chains) and compare it against the deployments. What do you think?

germartinez:
The purpose of having this config is to support the [EIP-3770](https://eips.ethereum.org/EIPS/eip-3770) (prefixes before the addresses) in the SDK for all the networks supported (the ones included in `safe-deployments`).

The config is updated when there is a new version of `safe-deployments`, but it needs to be something local in the code to be checked at runtime.

The current script (the one that makes the build fail right now) checks that the local config is up to date with the networks included in `safe-deployments`, otherwise the SDK whould be missing some prefixes for supported networks.

A different approach could be importing `ethereum-lists/chains` package, but that would increase the size of the Core SDK by a lot just to support a few networks. Knowing that `safe-deployments` is not released very often, I keep doing the update of this config manually.

katspaugh (commented):
Oh, I didn't know this file existed, we could re-use it in web-core as well, instead of [chains.ts](https://github.com/safe-global/web-core/blob/dev/src/config/chains.ts).

PR #304, opened 2022-11-29
Fix EthersAdapter documentation

germartinez:
## What it solves
The `EthersAdapter` class had its constructor updated, but not the documentation.
This pull request updates the docs in this regard.

It also updates some dependencies.


packages/safe-ethers-lib/README.md
```diff
@@ -45,15 +45,11 @@ const safeOwner = provider.getSigner(0)
...
+  signerOrProvider: safeOwner
 })
 ```
 
-In case the `ethAdapter` instance is only used to execute read-only methods the `signerAddress` property can be omitted.
-
-```js
-const readOnlyEthAdapter = new EthersAdapter({ ethers })
-```
+Depending on if the `ethAdapter` instance is used to execute or read-only methods the `signerOrProvider` property can be a `Signer` or a `Provider`.
```
katspaugh, line 52:
```suggestion
Depending on whether the `ethAdapter` instance is used to sign/execute transactions or just call read-only methods, the `signerOrProvider` property can be a `Signer` or a `Provider`.
```

katspaugh (approved):

iamacook (approved):

packages/safe-ethers-lib/README.md
germartinez: It looks perfect that way. Thank you!

PR #309, opened 2022-12-11
fix: Handle 0 safeTxGas when passed to createTransaction call

usame-algan:
## What it solves
Resolves [#1343](https://github.com/safe-global/web-core/issues/1343)

If a transaction is created with 0 safeTxGas in the web-core UI and another owner tries to sign it, the `createTransaction` call from the SDK will return a non 0 safeTxGas which causes issues with the existing txData (such as gasLimit estimation failing).

## How this PR fixes it

Narrow the condition to check for `undefined` so that a `safeTxGas` of 0 is caught by it.

### DIFF
--- a/packages/safe-core-sdk/tests/createTransaction.test.ts
+++ b/packages/safe-core-sdk/tests/createTransaction.test.ts
@@ -136,7 +136,7 @@ describe('Transactions creation', () => {
     )
 
     itif(safeVersionDeployed < '1.3.0')(
-      'should return a transaction with defined safeTxGas if safeVersion<1.3.0',
+      'should return a transaction with defined safeTxGas of 0 if safeVersion<1.3.0',
       async () => {
         const { accounts, contractNetworks } = await setupTests()
         const [account1, account2] = accounts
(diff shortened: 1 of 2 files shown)


packages/safe-core-sdk/tests/createTransaction.test.ts
```diff
@@ -147,7 +147,7 @@ describe('Transactions creation', () => {
           safeAddress: safe.address,
           contractNetworks
         })
-        const safeTxGas = 111
+        const safeTxGas = 0
```
mikhailxyz, line 150: Why this change? For this test it should be irrelevant if it's 0 or 111 ("should return a transaction with defined safeTxGas if safeVersion<1.3.0")

mikhailxyz: I think I get it but i'd a add a separate test for handling falsy values

usame-algan: Yeah, I was thinking its the edge value that caused the issue so rather have a test for it but I will add another.

mikhailxyz (approved):

mikhailxyz:
wait for a review from @germartinez cos he's the code owner :)

germartinez (approved):

PR #337, opened 2023-02-09
fix: allow EthersAdapter to correctly recognize Signer objects from o…

### LINKED ISSUE #336: Unable to use wrapped ethers signers from third-party libraries to create an EthersAdapter
## Description

Using a Signer object from an ethers.js library older than v5.7.0 to create an EthersAdapter results in the Signer being treated as a Provider. EthersAdapter seems to not recognize wrapped signers correctly and wrongly labeling them as providers only. An example is the [SignerWithAddress](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-ethers/src/signers.ts) wrapper provided by [`hardhat-ethers`](https://github.com/NomicFoundation/hardhat/tree/main/packages/hardhat-ethers) (which uses ethers.js v5.0.0)

## Environment

- Safe Core SDK version: Latest
- Safe contract version: Latest
- Environment:
  - non-browser

## Steps to reproduce

`package.json`
```json
"dependencies": {
    "hardhat": "^2.12.6",
    "@nomiclabs/hardhat-ethers": "^2.2.2",
    "@safe-global/safe-core-sdk": "^3.3.1",
    "@safe-global/safe-ethers-lib": "^1.9.1",
}
```

```ts
import { ethers } from 'hardhat';
import EthersAdapter from '@safe-global/safe-ethers-lib';

const [ owner ] = await ethers.getSigners();

const ethAdapter = new EthersAdapter({
  ethers,
  signerOrProvider: owner
});

// This should not be undefined
console.log(await ethAdapter.getSignerAddress());
```

This logs `undefined`, indicating `EthersAdapter` failing to recognize `owner` as a proper Signer object and labeling it as a `Provider` instead.

## Additional context

I found [this](https://github.com/safe-global/safe-core-sdk/blob/a8d5494daed4672e3980097a4b8cd8aaf9a2b835/packages/safe-ethers-lib/src/EthersAdapter.ts#L51) specific check to cause the issue. Changing it to check against the Signer class of the `ethers` parameter from the constructor signature resolves the issue.

MisterMard:
…lder ethers

## What it solves
Resolves #336

## How this PR fixes it
Checking whether `signerOrProvider` is Signer against the provided `ethers`'s Signer from the constructor parameters instead.

### DIFF
--- a/packages/safe-ethers-lib/src/EthersAdapter.ts
+++ b/packages/safe-ethers-lib/src/EthersAdapter.ts
@@ -48,7 +48,7 @@ class EthersAdapter implements EthAdapter {
       throw new Error('ethers property missing from options')
     }
     this.#ethers = ethers
-    const isSigner = signerOrProvider instanceof Signer
+    const isSigner = signerOrProvider instanceof ethers.Signer
     if (isSigner) {
       const signer = signerOrProvider as Signer
       if (!signer.provider) {


MisterMard:

germartinez:
Closing this after merging https://github.com/safe-global/safe-core-sdk/pull/373

PR #340, opened 2023-02-23
Fix: update safe-deployments

katspaugh:
## What it solves

The `safe-deployments` package was migrated to a new org + the latest version adds a new network.


packages/safe-core-sdk-utils/package.json
```diff
@@ -31,7 +31,6 @@
   ],
   "homepage": "https://github.com/safe-global/safe-core-sdk#readme",
   "devDependencies": {
-    "@gnosis.pm/safe-deployments": "1.19.0",
```
katspaugh, line 34: Does it really need to be only a dev dependency?

germartinez:
The `networks` object need to be updated here: `safe-core-sdk/packages/safe-core-sdk-utils/src/eip-3770/config.ts`

packages/safe-core-sdk-utils/package.json
germartinez: Inside the `safe-core-sdk-utils` package, yes

germartinez:
@katspaugh These chainIds should be added as well: 108, 4689, 9728, 43288, 56288

germartinez (approved):

PR #342, opened 2023-03-01
feat: add `decodeMultiSendData` function

### LINKED ISSUE #291: Decode MultiSend transactions into MetaTransactionData[]
Currently MultiSend transactions can only be encoded in the `safe-core-sdk`.
It would be a nice to have to allow decoding transactions in the `safe-core-sdk` as well.

Currently only this option to decode is available in the `safe-service-client`, but it does not return a `MetaTransactionData[]`:
```js
const decodedData = await safeService.decodeData(data)
```

Use this code from Manu as a reference:
```js
export const decodeMultiSendTxs = (data: string): MetaTransactionData[] => {
  const multiSendInterface = new ethers.utils.Interface([
    'function multiSend(bytes memory transactions) public payable ',
  ])
  const abiCoder = new ethers.utils.AbiCoder()
  // decode multiSend and remove '0x'
  let remainingData = multiSendInterface.decodeFunctionData('multiSend', data)[0].slice(2)

  const txs: MetaTransactionData[] = []
  while (remainingData.length > 0) {
    const txDataEncoded = ethers.utils.hexZeroPad(`0x${remainingData.slice(2, 170)}`, 32 * 3)
    const [txTo, txValue, txDataByteLength] = abiCoder.decode(['address', 'uint256', 'uint256'], txDataEncoded)
    remainingData = remainingData.slice(170)

    const dataLength = (txDataByteLength as BigNumber).toNumber() * 2
    let txData = `0x${remainingData.slice(0, dataLength)}`
    remainingData = remainingData.slice(dataLength)
    txs.push({
      to: txTo.toString(),
      value: txValue.toString(),
      data: txData,
      operation: 0,
    })
  }

  return txs
}

### PR DESCRIPTION
**Pull request by iamacook**
## What it solves

Resolves #291

## How this PR fixes it

This adds `decodeMultiSendData` which decodes the transactions of a `multiSend`.

### CODE REVIEW

File: packages/safe-core-sdk/src/utils/transactions/utils.ts [hunk 1]
```diff
@@ -80,3 +81,50 @@ function encodeMetaTransaction(tx: MetaTransactionData): string {
 export function encodeMultiSendData(txs: MetaTransactionData[]): string {
   return '0x' + txs.map((tx) => encodeMetaTransaction(tx)).join('')
 }
+
+export function decodeMultiSendData(data: string): MetaTransactionData[] {
+  // uint8 operation, address to, value uint265, dataLength uint256
+  const INDIVIDUAL_TX_DATA_LENGTH = 2 + 40 + 64 + 64
```
mikhailxyz, line 87: would be good to specify the units

packages/safe-core-sdk/src/utils/transactions/utils.ts
```diff
@@ -80,3 +82,35 @@ function encodeMetaTransaction(tx: MetaTransactionData): string {
...
+  const [decodedData] = multiSendInterface.decodeFunctionData('multiSend', encodedData)
+
+  const txs: MetaTransactionData[] = []
+
+  // Decode after 0x
+  let index = 2
+
+  while (index < decodedData.length) {
+    // As we are decoding hex encoded bytes calldata, each byte is represented by 2 chars
+    // uint8 operation, address to, value uint265, dataLength uint256
```
mikhailxyz, line 99:
```suggestion
    // uint8 operation, address to, value uint256, dataLength uint256
```

mikhailxyz (approved):

PR #343, opened 2023-03-02
Update collectibles endpoint to v2

dasanra:
## What it solves
- [BREAKING CHANGE] Upgrade the `collectibles` endpoint to v2 with pagination support.
- Fix a web3 types issue now that @types/web3 was deprecated with web3@v4 launch.

## How this PR fixes it
- Updated all endpoints to have versioning.
- [BREAKING CHANGE] Use collectibles v2 and add new properties to the endpoint
- Use fixed version of `@types/web3` until web3@v4 migration or can be changed by web3@v1 native types.


Uxio0 (approved):
Good job, thanks @dasanra

yagopv (approved):

Issue #359, opened 2023-03-20
Move the auth-kit package


dasanra:
## Overview

We need to add the `auth-kit` to the `safe-core-sdk` repository

## Requirements

- [x] Move `auth-kit` package to `safe-core-sdk` repository
- [x] Configure tests to run under the current action setup for `safe-core-sdk` repository
- [x] Set dependencies to use the kits existing on the same repository `safe-service-client` -> `api-kit`
- [x] Update repository references in docs, like the README

Issue #358, opened 2023-03-20
Move the onramp-kit package


dasanra:
## Overview

We need to add the `onramp-kit` to the `safe-core-sdk` repository

## Requirements

- [x] Move `onramp-kit` package to `safe-core-sdk` repository
- [x] Configure tests to run under the current action setup for `safe-core-sdk` repository
- [x] Update repository references in docs, like the README

PR #360, opened 2023-03-20
chore(auth-kit, onramp-kit): Add kits

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/359, https://github.com/safe-global/safe-core-sdk/issues/358

## How this PR fixes it
- [] Adding the kit from @safe-global/account-abstraction-sdk
- Configuring actions
- Configuting tests


package-lock.json
dasanra, line 1: We should use yarn.lock only

PR #372, opened 2023-03-31
feat(onramp-kit): Monerium integration

### LINKED ISSUE #376: Add Monerium support
## Overview

To continue evolving the SDK we need to add providers that enable improved UX flows. One interesting provider to allow on/off ramp is Monerium, that enables SEPA transfers directly from blockchain accounts on both directions

## Requirements

- [ ] Integrate Monerium SDK as a provider on the onramp kit.
  - If the onramp interface doesn't match Monerium needs evaluate to add as a different kit.
- [ ] Integrate the auth flow, where user giver permission to add to their account details
- [ ] Allow to create create a new account if it doesn't exist.
- [ ] Allow to view existing accounts
- [ ] Allow to get information from user transactions, so the app can display extra Monerium information on transaction details

yagopv:
## What it solves
Resolves #376 

## How this PR fixes it
We are creating a new pack allowing users to integrate the Safe with Monerium. An example integration can be explored [here](https://github.com/safe-global/safe-core-sdk/tree/main/packages/onramp-kit/example/client)

We are implementing our [common interface](https://github.com/safe-global/safe-core-sdk/blob/f2e8e82d88d815d7b278f605a125f4cfb2816020/packages/onramp-kit/src/types.ts#L10) providing the following methods:

- `init(MoneriumPack)` - Allow to initialize the kit with the `MoneriumPack`. We are going to provide the clientId and choose the Monerium environment to work with (sandbox or production)
- `open(MoneriumOpenOptions)` - We are going to start the Monerium flow in 3 ways:
    - Redirecting to the Monerium website and let the user login and provide access to the resources
    - By providing an authCode in the open options to exchange and get access to the resources
    - By providing a refreshToken in the open options to authenticate and get access to the resources
- `subscribe(event, handler) / unsubscribe(event)` - Once authenticated this methods allow the user to establish a socket connection with the Monerium servers and listen for order status changes
- `close()` - To cleanup the connection with Monerium

## How to test it
- [Follow the instructions](https://github.com/safe-global/safe-core-sdk/blob/feat/monerium-integration/packages/onramp-kit/example/client/README.md) to build the monorepo and the example
- Create an account in [Monerium sandbox](https://sandbox.monerium.dev/) and then [create an application](https://monerium.dev/docs/getting-started/create-app) for obtaining the clientId. Set the clientId in the .env file

**Use Cases**
1) Signin/Signup and link a 1/1 Safe. Should bind the Safe and autoexecute transactions as only the Web3Auth signer is required (Remember to send funds to the owner address as it will need them to autoexecute the txs)
2) Signin/Signup and link a 2/X Safe. Should bind the Safe and propose the signMessage transaction. Other signers should go to the Safe to sign
3) Send transactions (default to 1 EURe) between Monerium sandbox accounts and check they arrive. Same as 1) and 2) txs should autoexecute or require additional signers (In the Safe UI)

Read [integration notes](https://www.notion.so/Monerium-integration-103cdd3c3d13471e8ef6df29cfc12456
) and attached videos to check how the example behaves


packages/onramp-kit/src/SafeOnRampKit.ts
```diff
@@ -29,13 +30,14 @@ export class SafeOnRampKit<TPack extends SafeOnRampPack<TPack>> {
    * @throws Error if the pack is not defined
    */
   static async init<TPack extends SafeOnRampPack<TPack>>(
-    pack: TPack
+    pack: TPack,
+    safeSdk?: Safe
```
dasanra, line 34: Missing parameter description at JS doc

packages/onramp-kit/example/client/src/AuthContext.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { createContext, useState, useEffect } from 'react'
+import { Web3AuthOptions } from '@web3auth/modal'
+import { CHAIN_NAMESPACES, SafeEventEmitterProvider, WALLET_ADAPTERS } from '@web3auth/base'
+import { OpenloginAdapter } from '@web3auth/openlogin-adapter'
+
+// TODO: Change to @safe-global/auth-kit when next version published
+import { SafeAuthKit, Web3AuthModalPack, SafeAuthSignInData } from '../../../../auth-kit/src/index'
```
dasanra, line 7: Should be published now

packages/onramp-kit/src/packs/monerium/signatures.ts
```diff
@@ -0,0 +1,13 @@
+import { ethers } from 'ethers'
+
+const MAGIC_VALUE = '0x1626ba7e'
+const MAGIC_VALUE_BYTES = '0x20c13b0b'
```
dasanra, line 4: Where does these magic values come from?

I'm missing some comment with context

packages/onramp-kit/src/types.ts
```diff
@@ -1,14 +1,23 @@
...
+  SafeMoneriumClient,
+  MoneriumEvent,
+  MoneriumEventListener,
+  MoneriumOpenOptions
+} from './packs/monerium'
 
 // The new packs must implement this interface
 export interface SafeOnRampPack<TPack> {
-  init(): Promise<void>
+  init(safeSdk?: Safe): Promise<void>
```
dasanra, line 20: This is optional for Stripe kit, but mandatory for Monerium. I would type it like you did in the `open` function below. Use an `options` object that will have different type depending on the provided pack.

packages/onramp-kit/src/SafeOnRampKit.ts
yagopv: done

packages/onramp-kit/src/packs/monerium/signatures.ts
yagopv: done

packages/onramp-kit/src/types.ts
yagopv: done

packages/onramp-kit/example/client/src/AuthContext.tsx
yagopv: Removed

packages/onramp-kit/src/packs/stripe/StripePack.ts
```diff
@@ -51,7 +51,7 @@ export class StripePack implements SafeOnRampPack<StripePack> {
     sessionId,
     defaultOptions
   }: StripeOpenOptions): Promise<StripeSession> {
-    if (!this.#stripeOnRamp) throw new Error('StripeOnRamp is not initialized')
+    if (!this.#client) throw new Error('StripeOnRamp is not initialized')
```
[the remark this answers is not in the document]
yagopv: Updated

packages/onramp-kit/src/packs/monerium/SafeMoneriumClient.ts
```diff
@@ -0,0 +1,294 @@
...
+        // @ts-expect-error - dataDecoded should have the method property
+        tx?.dataDecoded?.method === 'signMessage' &&
+        // @ts-expect-error - dataDecoded should have the parameters array
+        tx?.dataDecoded?.parameters[0]?.value === ethers.utils.hashMessage(message)
+      ) {
+        return true
+      }
+
+      return false
+    })
```
dasanra, line 93:
```suggestion
    const isMessagePending = pendingTransactions.results.some((tx) => (
        // @ts-expect-error - dataDecoded should have the method property
        tx?.dataDecoded?.method === 'signMessage' &&
        // @ts-expect-error - dataDecoded should have the parameters array
        tx?.dataDecoded?.parameters[0]?.value === ethers.utils.hashMessage(message)
   ))
```

dasanra (approved):

PR #382, opened 2023-04-10
Duplicated code cleanup

dasanra:
## What it solves
Some duplicated code still existed on the account-abstraction-kit. Also we were doing additional checks when using Gelato 1Balance that are not needed

## How this PR fixes it
Remove duplicated code and use from protocol-kit exported functions
Code cleanup to select when transaction is sponsored or using sync balance and prevent aditional calculations not needed for 1Balance


germartinez (approved):

germartinez:
Tests need to be updated

germartinez (approved):

PR #393, opened 2023-04-12
refactor(protocol-kit): add predict Safe address method as a util function

### LINKED ISSUE #355: Expose counterfactual deployment methods from Protocol Kit
## Overview

Currently there are methods to predict the Safe address based on the deployment configuration. It's created behind the safeFactory class which makes it more difficult to use as you need to instance the full class even you not use it.

## Proposed solution

- [x] Extract this function so it can be used as an util, not only inside a class. Tha way it can be reused for other packages.
- [x] Document how to use this functionality.
- [x] Use this feature with the relay to deploy a 1:1 Safe on an example.

DaniSomoza:
## What it solves
Resolves #355

## How this PR fixes it

Added `predictSafeAddress` method as a util function

### DIFF
--- a/packages/protocol-kit/src/contracts/utils.ts
+++ b/packages/protocol-kit/src/contracts/utils.ts
@@ -62,16 +62,30 @@ export async function encodeSetupCallData({
     ])
   }
 
-  const fallbackHandlerAddress =
-    fallbackHandler ||
-    (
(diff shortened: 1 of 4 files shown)


DaniSomoza:
Implemented in #415

PR #404, opened 2023-04-17
fix!: use type string for uint256 parameters

dasanra:
## What it solves
Solves #410
Javascript `number` is only able to manage safely numbers up to 2^53-1. Some parameters used on the Safe smart contracts allow to use `uint256` solidity type numbers. That would mean that those parameters can use 2^256-1 values. This would cause overflow on those parameters that are typed with Javascript `number`

## How this PR fixes it
There are two ways of solving this:
 - Using BigNumber library
 - Using strings to send those values.
 
 As we want to support Web3 and Ethers it's difficult to use a BigNumber library, as they use a different one that would make types more difficult to match. To handle values without depending on a library and loosing precision then we should use strings when sending those values, and converting internally to BigNumber when we need to modify them.


packages/protocol-kit/tests/e2e/createTransaction.test.ts
```diff
@@ -73,14 +82,14 @@ describe('Transactions creation', () => {
...
-          gasPrice: 123
+          gasPrice: '123'
         }
         const safeTxData = await standardizeSafeTransactionData(
           safeSdk.getContractManager().safeContract,
           ethAdapter,
           txDataPartial
         )
-        chai.expect(safeTxData.safeTxGas).to.be.gt(0)
+        chai.expect(Number(safeTxData.safeTxGas)).to.be.gt(0)
```
germartinez, line 92: Should we use `BigNumber` here in this test?

packages/protocol-kit/tests/e2e/createTransaction.test.ts
```diff
@@ -134,7 +143,7 @@ describe('Transactions creation', () => {
...
```
germartinez, line 146: Should we use `BigNumber` here in this test?

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -309,7 +309,7 @@ describe('Off-chain signatures', () => {
         isSuccessful: false,
         ethGasPrice: '',
         gasUsed: 12345,
-        fee: 12345,
+        fee: '12345',
```
germartinez, line 312: `fee` is defined as an integer here: https://safe-transaction-mainnet.safe.global/
(same as `baseGas` and `safeTxGas` in lines 296-297)

packages/safe-core-sdk-types/src/ethereumLibs/EthAdapter.ts
```diff
@@ -19,8 +19,8 @@ export interface EthAdapterTransaction {
   from: string
   data: string
   value?: string
-  gasPrice?: number
-  gasLimit?: number
+  gasPrice?: number | string
+  gasLimit?: number | string
```
germartinez, line 23: Why is `number` allowed?
I guess same happens with `maxFeePerGas` and `maxPriorityFeePerGas` in lines 24-25

packages/protocol-kit/src/adapters/web3/Web3Adapter.ts
```diff
@@ -64,6 +64,7 @@ class Web3Adapter implements EthAdapter {
     const balance = defaultBlock
       ? await this.#web3.eth.getBalance(address, defaultBlock)
       : await this.#web3.eth.getBalance(address)
+    // FIXME Web3 Adapter is forced to return an Ethers type
```
germartinez, line 67: We should probably create a new ticket for this so we don't forget

packages/protocol-kit/src/Safe.ts
```diff
@@ -794,7 +794,7 @@ class Safe {
       operation: serviceTransactionResponse.operation,
       safeTxGas: serviceTransactionResponse.safeTxGas,
       baseGas: serviceTransactionResponse.baseGas,
-      gasPrice: Number(serviceTransactionResponse.gasPrice),
+      gasPrice: serviceTransactionResponse.gasPrice,
```
germartinez, line 797: According to https://safe-transaction-mainnet.safe.global/, `baseGas` and `safeTxGas` in lines 795-796 are integers so we should convert them to strings if that is true

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
dasanra: imagen

We can't just trust swagger 🙈

packages/safe-core-sdk-types/src/types.ts
```diff
@@ -147,8 +146,8 @@ export type SafeMultisigTransactionResponse = {
   readonly data?: string
   readonly operation: number
   readonly gasToken: string
-  readonly safeTxGas: number
-  readonly baseGas: number
+  readonly safeTxGas: string
```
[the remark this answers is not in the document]
dasanra: I updated 149 and 150. 165 was wrong in the `api-kit` type

packages/safe-core-sdk-types/src/ethereumLibs/EthAdapter.ts
dasanra: As this type is used always as input typing for the providers I left it a bit more flexible. Both libraries would admit
```undefined | number | string | BN/BigNumber```

We are already removing the BN/BigNumber possibility.

packages/protocol-kit/src/adapters/web3/Web3Adapter.ts
dasanra: https://github.com/safe-global/safe-core-sdk/issues/413

germartinez (approved):

mmv08:
Another alternative would be using JS' BigInt (just saying)

dasanra:
@mmv08 we also considered that, but would need deeper changes and more effort as Ethers v5 it's still not compatible and Web3 neither

PR #408, opened 2023-04-18
fix: TokenInfoListResponse results type

sirvar:
TokenInfoListResponse results is a circular reference. As per the API docs, it should be TokenInfoResponse.

## What it solves
Resolves an incorrect type in `api-kit`

## How this PR fixes it
Properly define TokenInfoListResponse type


sirvar:

sirvar:
recheck

dasanra (approved):
Looks good! Thank you for your fix!

PR #416, opened 2023-04-25
chore: Rename adapters to packs

yagopv:
## What it solves
- We are going to rename the current adapters in the kits to being suffixed with `Pack` instead:
    - GelatoAdapter → GelatoPack
    - Web3AuthAdapter → Web3AuthPack
    - StripeAdapter → StripePack
- So the basic snippet to use any kit should change to this:

```typescript
const web3AuthPack = new Web3AuthPack(options, [openloginAdapter], modalConfig)

const safeAuthKit = await SafeAuthKit.init(web3AuthPack, {

  txServiceUrl: 'https://safe-transaction-url.safe.global'

})
```


packages/onramp-kit/src/types.ts
```diff
@@ -1,28 +1,28 @@
...
-// When creating new adapters these types should be updated:
+// When creating new packs these types should be updated:
 // e.g.:
 // export type SafeOnRampOpenOptions<T> =
-//    T extends StripeAdapter ? StripeOpenOptions :
-//    T extends FooAdapter ? FooOpenOptions :
-//    T extends BarAdapter ? BarOpenOptions :
+//    T extends StripePack ? StripeOpenOptions :
+//    T extends FooPack ? FooOpenOptions :
+//    T extends BarPack ? BarOpenOptions :
```
germartinez, line 23: According to lines 25-28 it should be renamed here as well
```suggestion
//    TPack extends StripePack ? StripeOpenOptions :
//    TPack extends FooPack ? FooOpenOptions :
//    TPack extends BarPack ? BarOpenOptions :
```

packages/auth-kit/src/SafeAuthKit.ts
```diff
@@ -39,16 +37,16 @@ export class SafeAuthKit<TAdapter extends SafeAuthAdapter<TAdapter>>
    * @returns A SafeAuthKit instance
    * @throws Error if the provider type is not supported
    */
-  static async init<T extends SafeAuthAdapter<T>>(
-    adapter: T,
+  static async init<T extends SafeAuthPack<T>>(
+    pack: T,
```
germartinez, line 41: Maybe `TPack`?
```suggestion
  static async init<TPack extends SafeAuthPack<TPack>>(
    pack: TPack,
```

germartinez (approved):

dasanra (approved):

PR #420, opened 2023-04-27
feat(auth-kit): Expose the getUserInfo() method from web3auth

yagopv:
## What it solves
It allow users to use the [`getUserInfo()`](https://web3auth.io/docs/pnp/features/server-side-verification/social-login-users#getuserinfo) method from web3 auth

It expose it through the common interface as new providers can expose this information as well with their own methods

NOTE: It fix a bug where the provider was not initialised if i try to make a page refresh

### DIFF
--- a/packages/auth-kit/src/packs/web3auth/Web3AuthAdapter.ts
+++ b/packages/auth-kit/src/packs/web3auth/Web3AuthAdapter.ts
@@ -44,9 +44,9 @@ export class Web3AuthAdapter implements SafeAuthAdapter<Web3AuthAdapter> {
 
       this.#adapters?.forEach((adapter) => this.web3authInstance?.configureAdapter(adapter))
 
-      this.provider = this.web3authInstance.provider
-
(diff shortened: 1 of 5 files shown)


packages/auth-kit/src/packs/web3auth/Web3AuthAdapter.ts
```diff
@@ -56,22 +56,40 @@ export class Web3AuthAdapter implements SafeAuthAdapter<Web3AuthAdapter> {
...
+  /**
+   * Get authenticated user information
+   * @returns The user info
+   */
+  async getUserInfo(): Promise<Partial<UserInfo>> {
+    if (!this.web3authInstance) {
+      throw new Error('Web3AuthAdapter is not initialized')
+    }
+
+    const userInfo = await this.web3authInstance?.getUserInfo()
```
germartinez, line 88: This should not be needed because of the previous check
```suggestion
    const userInfo = await this.web3authInstance.getUserInfo()
```

germartinez (approved):

dasanra (approved):

PR #430, opened 2023-05-10
fix(protocol-kit): safeDeploymentConfig is an optional parameter in PredictedSafeProps

### LINKED ISSUE #426: safeDeploymentConfig property is mandatory
## Description

safeDeploymentConfig object is mandatory when initializing the safeSDK predicting the Safe address but all it's properties are optional. If you don't want to set any of the properties you still need to set an empty object.

## Environment

- Package: protocol-kit
- Package version: 1.0.0

## Expected result
Is not necessary to set an empty object if no need to set any of the properties

## More context
https://github.com/safe-global/safe-core-sdk/blob/f2e8e82d88d815d7b278f605a125f4cfb2816020/packages/protocol-kit/src/types/index.ts#L22-L29

DaniSomoza:
## What it solves
Resolves #426

## How this PR fixes it

`safeDeploymentConfig` is an optional parameter in the `PredictSafeProps` type

```
export interface PredictedSafeProps {
  safeAccountConfig: SafeAccountConfig
  safeDeploymentConfig?: SafeDeploymentConfig
}
```

### DIFF
--- a/packages/protocol-kit/src/contracts/utils.ts
+++ b/packages/protocol-kit/src/contracts/utils.ts
@@ -20,7 +20,7 @@ import { SAFE_LAST_VERSION } from '@safe-global/protocol-kit/contracts/config'
 export const PREDETERMINED_SALT_NONCE =
   '0xb1073742015cbcf5a3a4d9d1ae33ecf619439710b89475f92e2abd2117e90f90'
 
-export interface PredictSafeProps {
+export interface PredictSafeAddressProps {
   ethAdapter: EthAdapter
   safeAccountConfig: SafeAccountConfig
   safeDeploymentConfig?: SafeDeploymentConfig
@@ -108,7 +108,7 @@ export async function predictSafeAddress({
   safeDeploymentConfig = {},
   isL1SafeMasterCopy = false,
   customContracts
-}: PredictSafeProps): Promise<string> {
+}: PredictSafeAddressProps): Promise<string> {
   validateSafeAccountConfig(safeAccountConfig)
   validateSafeDeploymentConfig(safeDeploymentConfig)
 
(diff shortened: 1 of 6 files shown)


packages/protocol-kit/src/contracts/utils.ts
```diff
@@ -23,7 +23,7 @@ export const PREDETERMINED_SALT_NONCE =
 export interface PredictSafeProps {
```
germartinez, line 23: Renaming this type would help to differentiate it more from `PredictedSafeProps`. Also the name of the funciton is `predictedSafeAddress()`, so I think it makes sense.
```suggestion
export interface PredictSafeAddressProps {
```

packages/protocol-kit/tests/e2e/utilsContracts.test.ts
```diff
@@ -453,5 +454,41 @@ describe('Contract utils', () => {
       chai.expect(secondPredictedSafeAddress).to.be.equal(expectedSafeAddress)
       chai.expect(thirdPredictedSafeAddress).to.be.equal(expectedSafeAddress)
     })
+
+    itif(safeVersionDeployed > '1.0.0')(
```
germartinez, line 458: Beautiful test. Just checking this check is done because in case `safeDeploymentConfig` is not passed, the default version will be (currently) `v1.3.0` and its use would break with the `v1.0.0` contracts deployed by Hardhat when using `customContracts` variable

germartinez: when calling `setup` method in the contract, because of the number or parameters of the function

germartinez: Okay, what I was thinking. All good then!

packages/protocol-kit/src/contracts/utils.ts
DaniSomoza: done!

germartinez (approved):

PR #431, opened 2023-05-11
fix(account-abstraction-kit): use the protocol kit

DaniSomoza:
## What it solves
Resolves:

- fix the `relayTransaction` function in the `account-abstraction-kit`
- fix the initialization of the `account-abstraction-kit` when the Safe is not deployed.

## How this PR fixes it

Adding a instance of the protocol kit instead of using the Safe contract instance directly.

https://github.com/safe-global/safe-core-sdk/blob/1671f01bf9dad76416bc64ed0450465a040d9c69/packages/account-abstraction-kit/src/AccountAbstraction.ts#L31-L36

### DIFF
--- a/packages/account-abstraction-kit/src/AccountAbstraction.ts
+++ b/packages/account-abstraction-kit/src/AccountAbstraction.ts
@@ -18,7 +18,6 @@ import Safe, {
 } from '@safe-global/protocol-kit'
 import { RelayPack } from '@safe-global/relay-kit'
 import {
-  GnosisSafeContract,
   GnosisSafeProxyFactoryContract,
   MetaTransactionData,
   MetaTransactionOptions,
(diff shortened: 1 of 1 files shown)


dasanra (approved):

packages/account-abstraction-kit/src/AccountAbstraction.ts
```diff
@@ -76,11 +75,22 @@ class AccountAbstraction {
       safeDeploymentConfig
     })
 
-    this.#safeContract = await getSafeContract({
-      ethAdapter: this.#ethAdapter,
-      safeVersion,
-      customSafeAddress: safeAddress
-    })
+    try {
+      await getSafeContract({
```
germartinez, line 79: This should be removed

PR #432, opened 2023-05-11
fix(account-abstraction-kit): remove unnecessary line

germartinez:
## What it solves
Removes an unnecessary line

### DIFF
--- a/packages/account-abstraction-kit/src/AccountAbstraction.ts
+++ b/packages/account-abstraction-kit/src/AccountAbstraction.ts
@@ -76,12 +76,6 @@ class AccountAbstraction {
     })
 
     try {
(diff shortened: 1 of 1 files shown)


packages/account-abstraction-kit/src/AccountAbstraction.ts
```diff
@@ -76,12 +76,6 @@ class AccountAbstraction {
     })
 
     try {
-      await getSafeContract({
```
DaniSomoza, line 79: This line is necessary, it is a workaround to detect if the Safe is deployed

germartinez: Okay, I see. I'll close the PR

PR #436, opened 2023-05-11
fix(protocol-kit): fix standardizeSafeTransactionData for predictedSafe flow

### LINKED ISSUE #427: Can't create a Safe transaction if the Safe version is not set explicitly in predict config
## Description

Is not possible to create a Safe transaction using an SDK instance initialized via configuration to predict the Safe address. The `standardizeSafeTransactionData` method check for this value in the contract manager where it isn't set, not allowing to continue if is not explicitly set on the configuration object. The safeSDK is able to resolve the expected Safe version, so it should be possible to continue without providing this parameter.

## Environment

- Package: `protocol-kit` 
- Package version: .0.0 

## Expected result

I should be able to create a Safe transaction even if there is not explicit Safe version in the safeSDK predict config object.

## Additional context

https://github.com/safe-global/safe-core-sdk/blob/f2e8e82d88d815d7b278f605a125f4cfb2816020/packages/protocol-kit/src/utils/transactions/utils.ts#L54-L62

DaniSomoza:
## What it solves
Resolves #427

I should be able to create a Safe transaction even if there is not explicit Safe version in the safeSDK predict config object.

## How this PR fixes it

```
  if (predictedSafe) {
    safeVersion = predictedSafe?.safeDeploymentConfig?.safeVersion || SAFE_LAST_VERSION
```

https://github.com/safe-global/safe-core-sdk/blob/f2e8e82d88d815d7b278f605a125f4cfb2816020/packages/protocol-kit/src/utils/transactions/utils.ts#L54-L62

### DIFF
--- a/packages/protocol-kit/src/utils/transactions/utils.ts
+++ b/packages/protocol-kit/src/utils/transactions/utils.ts
@@ -53,7 +53,7 @@ export async function standardizeSafeTransactionData({
   }
 
   let safeVersion: SafeVersion
-  if (!!predictedSafe) {
+  if (predictedSafe) {
     safeVersion = predictedSafe?.safeDeploymentConfig?.safeVersion || SAFE_LAST_VERSION
   } else {
     if (!safeContract) {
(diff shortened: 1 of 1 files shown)


packages/protocol-kit/src/utils/transactions/utils.ts
```diff
@@ -52,8 +53,8 @@ export async function standardizeSafeTransactionData({
   }
 
   let safeVersion: SafeVersion
-  if (predictedSafe?.safeDeploymentConfig?.safeVersion) {
-    safeVersion = predictedSafe?.safeDeploymentConfig.safeVersion
+  if (!!predictedSafe) {
```
germartinez, line 56: Why not `if (predictedSafe)`?
I understand that `!!` converts into a boolean with the same value, but I also find it a little bit more difficult to read than just having `if (predictedSafe)`

DaniSomoza: done!

germartinez (approved):

PR #445, opened 2023-05-19
feat(protocol-kit): Add memoization to the predictSafeAddress util function to enhance its performance

DaniSomoza:
## What it solves

Add [memoization](https://en.wikipedia.org/wiki/Memoization) to the `predictSafeAddress` util function to enhance its performance.

This new version of the `predictSafeAddress` is approximately 50,000 times faster than the old one.

## How this PR fixes it

Added a new generic `createMemoizedFunction` function to memoize contract instance fetching in our `predictSafeAddress` function to enhance its performance.

```
export function createMemoizedFunction<T extends (...args: Parameters<T>) => ReturnType<T>>(
  callback: T,
  cache: Record<string, ReturnType<T>> = {}
) {
  return (...args: Parameters<T>): ReturnType<T> => {
    const key = JSON.stringify(args)

    cache[key] = cache[key] || callback(...args)

    return cache[key]
  }
}
```

- Old version of `predictSafeAddress` function can generate only 1 Safe address each 6 seconds.
- New version of `predictSafeAddress` function can generate more than 6.000 Safe Addresses each second.

Based on that, this new version of the `predictSafeAddress` is approximately 50,000 times faster than the old one.

See this demo (left old version, right new efficient version):

https://github.com/safe-global/safe-core-sdk/assets/26763673/fba18505-a670-4240-9775-4f1450cd959d


README.md
```diff
@@ -8,21 +8,21 @@ Software developer tools that facilitate the interaction with the Safe [contract
 
 ## Guides
 
-| Title | Description |
+| Title                                                                                                                           | Description                                                                                                                                                                                                          |
```
yagopv, line 11: This is intentional?

README.md
```diff
@@ -8,21 +8,21 @@ Software developer tools that facilitate the interaction with the Safe [contract
...
 | ------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
yagopv, line 12: Same ⬆️

README.md
```diff
@@ -8,21 +8,21 @@ Software developer tools that facilitate the interaction with the Safe [contract
...
-| [safe-ethers-adapters](https://github.com/safe-global/safe-core-sdk/tree/main/packages/safe-ethers-adapters) | [![NPM Version](https://badge.fury.io/js/%40safe-global%2Fsafe-ethers-adapters.svg)](https://badge.fury.io/js/%40safe-global%2Fsafe-ethers-adapters) | [Ethers](https://docs.ethers.io/v5/single-page/) adapter that facilitates the interaction with the [Safe Services](https://github.com/safe-global/safe-transaction-service) |
+| Package                                                                                                      |                                                                       Release                                                                        | Description                                                                                                                                                                        |
```
yagopv, line 17: Intentional?

yagopv (approved):

README.md
germartinez: This is an autoformatter, like in line 12. I don't like how it does it either, would be good if we could avoid it in the tables inside all the repo

DaniSomoza: removed!

README.md
DaniSomoza: removed!

dasanra (approved):

PR #450, opened 2023-05-26
feat(protocol-kit): smart contract signatures (EIP-1271)

### LINKED ISSUE #407: Add support for smart contract signatures
## Context / issue
Currently it is not possible to create dynamic EIP-1271 signatures through the safe-core-sdk.
These are important for smart contract signatures such as Safes or other smart contract wallets as Safe owners.

## Proposed solution
We should add a separate `SmartContractSignature` or `EIP1271Signature` implementation of `SafeSignature`.

Therefore we should add back the `offset` parameter to the `staticPart` function in order to easily provide the pointer to the start of the `dynamicPart`.

## Alternatives
Alternatively we could leave the parameters of the functions as they are and implement the EIP1271 Signature like this:
```ts
export class EIP1271Signature implements SafeSignature {
  constructor(public signer: string, public data: string, private offset: number) {}

  staticPart(): string {
    return hexZeroPad(this.signer, 32) + hexZeroPad(hexlify(this.offset), 32).slice(2) + '00'
  }
  dynamicPart(): string {
    const byteLength = hexZeroPad(hexlify(this.data.slice(2).length / 2), 32)
    return (byteLength + this.data.slice(2)).slice(2)
  }
}
```

Then the dapps would have to keep track of the current signature and pass in the offset in the constructor.

## Additional context
With this feature we could easily support all kinds of smart contract wallet in our Safe{Wallet}

schmanu:
## What it solves
Currently we do not support off-chain signing for smart contract owners using EIP-1271.

Resolves #407 

## How this PR fixes it
- Adds new dynamic `SmartContractSafeSignature`
- When adding a signature to a transaction we check if the signer address is a smart contract and if it is we use the new signature for smart contracts

## Todo
- [ ] Extend e2e test with a test for nested Safes
- [ ] Check `isValidSignature` before adding the signature?
- [ ] Copy new signatures correctly
- [ ] Build SafeTransaction from transaction service response correctly

### DIFF
--- a/packages/protocol-kit/README.md
+++ b/packages/protocol-kit/README.md
@@ -640,7 +640,7 @@ const txHash = await safeSdk.getTransactionHash(safeTransaction)
 const signature = await safeSdk.signTransactionHash(txHash)
 ```
 
-### signTypedData
+### signTransactionTypedData
 
 Signs a transaction according to the EIP-712 using the current signer account.
 
@@ -649,7 +649,7 @@ const safeTransactionData: SafeTransactionDataPartial = {
   // ...
 }
 const safeTransaction = await safeSdk.createTransaction({ safeTransactionData })
-const signature = await safeSdk.signTypedData(safeTransaction)
+const signature = await safeSdk.signTransactionTypedData(safeTransaction)
 ```
 
 ### signTransaction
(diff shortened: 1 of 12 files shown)


dasanra:
Implemented in https://github.com/safe-global/safe-core-sdk/pull/556

PR #470, opened 2023-06-29
feat: Coverage comments

yagopv:
## What it solves

Add specific sections in `test.yml` to upload the code coverage reports to Coveralls

https://coveralls.io/github/safe-global/safe-core-sdk


yagopv:
the protocol-kit is a little different because we only have 2 unit tests specs, the others are e2e specs using ganache and are being executed in a different way (and other action). We need a way to generate a report with the same format as the others as well so is not as straightforward as the others

Probably we need to proper unit test the packages as well using jest

germartinez (approved):

PR #471, opened 2023-07-04
chore(safe-ethers-adapters): Remove safe-ethers-adapters package

germartinez:
## What it solves
This package is not maintained since a very long time ago.
Also it is not very related to the new approach we have with the kits.
It might make sense to move it to https://github.com/5afe and try to keep this repo just with the minimum required number of packages, knowing than the number of kits will increase in the future.


yagopv (approved):
Always a pleasure to delete things :)

PR #477, opened 2023-07-12
feat(api-kit): add pagination support to getAllTransactions

tukantje:
## What it solves
`getAllTransactions` method of Safe Api Kit is a paginated endpoint, however, when using the Api Kit we are left with only the first page's result. Fetching rest needs to be done by hand.

## How this PR fixes it
This PR adds two new fields to `getAllTransactions` result, which are added based on whether or not next / previous pages exist;
* `getNext` fetches the next page, using the URL returned from the previous response.
* `getPrevious` fetches the previous page, using the URL returned from the previous response.


tukantje:

shoom3301 (approved):

PR #414, opened 2023-04-20
feat(relay-kit): make gas limit optional

dasanra:
## What it solves
Resolves #362

### Features added
- `GasLimit` optional in the `relay-kit`
- Full Counterfactual deployment feature added to the `relay-kit`.
- Improved usability of the `relay-kit` (no need to manually encode transactions by developers)

New Syntax

```javascript
 
  // dump transfer as an example
  const dumpTransfer = {
    to: '0x123', // to address
    data: '0x',
    value: '1000000000000000' // 0.001 ETH
  }

  // batch to be executed by Gelato
  const transactions: MetaTransactionData[] = [
     dumpTransfer
  ]

  // we create the Gelato relay pack instance
  const relayPack = new GelatoRelayPack()

  // we create the relayed transaction
  const safeTransaction = await relayPack.createRelayedTransaction({ safe, transactions })

  // we sign the relayed transaction with all the owners
  const signedSafeTransaction = await signTransactionWithAllOwners(safe, safeTransaction)

  // we execute the relayed transaction with all the signatures
  const { taskId } = await relayPack.executeRelayTransaction(signedSafeTransaction, safe)
```

## How this PR fixes it

Added Functions:

### `Protocol-kit`

- [`getEncodedTransaction`](https://github.com/safe-global/safe-core-sdk/pull/414/files#diff-e61078c05e6c2c9d498623afd936ffe855d05a132d757f11f4ad047e030c77a4R1024) in the `protocol-kit`: Returns the Safe Transaction encoded (as a string).
- [`wrapSafeTransactionIntoDeploymentBatch`](https://github.com/safe-global/safe-core-sdk/pull/414/files#diff-e61078c05e6c2c9d498623afd936ffe855d05a132d757f11f4ad047e030c77a4R1060) in the `protocol-kit`: Wraps a Safe transaction into a Safe deployment batch.
- [`createSafeDeploymentTransaction`](https://github.com/safe-global/safe-core-sdk/pull/414/files#diff-e61078c05e6c2c9d498623afd936ffe855d05a132d757f11f4ad047e030c77a4R1116) in the `protocol-kit`:  Creates a Safe deployment transaction of the Safe.
- [`createTransactionBatch`](https://github.com/safe-global/safe-core-sdk/pull/414/files#diff-e61078c05e6c2c9d498623afd936ffe855d05a132d757f11f4ad047e030c77a4R1179) in the `protocol-kit`: This function creates a batch transaction of the provided Safe transactions using the MultiSend contract.

[...]

PR #496, opened 2023-07-31
feat(relay-kit): Add separate ERC20 transfer for fee payment with non-standard ERC20 tokens

DaniSomoza:
Add separate ERC20 transfer for fee payment with non-standard ERC20 tokens

## How this PR fixes it

Added a new `createTransactionWithTransfer` function that uses a separate ERC20 transfer to pay the fees to the Gelato relayer.


dasanra:
Fix this naming issue in this PR
https://github.com/safe-global/safe-core-sdk/pull/414#pullrequestreview-1554553936

packages/protocol-kit/src/utils/erc-20/index.ts
```diff
@@ -0,0 +1,90 @@
...
+): Promise<boolean> {
+  const isNativeToken = gasToken === ZERO_ADDRESS
+
+  if (isNativeToken) {
+    return true
+  }
+
+  // Only ERC20 tokens with the standard 18 decimals are compatible
+  const gasTokenDecimals = await getERC20Decimals(gasToken, safe)
+  const isStandardERC20Token = gasTokenDecimals === STANDARD_ERC20_DECIMALS
```
dasanra, line 62: I would call this `hasTokenStandardERC20Decimals` because checking if it's a standard ERC20 would involve more things than just checking it has 18 decimals

packages/protocol-kit/src/utils/erc-20/index.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Creates a transaction object to perform a transfer of a specified amount of ERC-20 tokens to a given address.
+ *
+ * @export
+ * @param {string} tokenAddress - The address of the ERC-20 token.
+ * @param {string} toAddress - The address to which the tokens should be transferred.
+ * @param {string} amount - The amount of tokens to transfer.
+ * @returns {Transaction} Returns a transaction object that represents the transfer.
+ */
+export function createERC20tokenTransferTransaction(
```
dasanra, line 76: There is a typo here, this can cause confusion when using it in other sections as it's an exported function name

`t` from `token` should be capital

packages/relay-kit/src/packs/gelato/GelatoRelayPack.ts
```diff
@@ -69,7 +70,7 @@ export class GelatoRelayPack implements RelayPack {
     onlyCalls = false,
     options = {}
   }: CreateTransactionProps): Promise<SafeTransaction> {
-    const { gasLimit, isSponsored = false } = options
+    const { isSponsored = false } = options
     const nonce = await safe.getNonce()
     const chainId = await safe.getChainId()
```
dasanra, line 75: We also execute this two calls in the helper functions when transaction is not sponsored, so either we set this inside the if (isSponsored) or we pass them as parameter to the HandlePayment and WithTransfer functions

dasanra (approved):

PR #509, opened 2023-08-16
fix(protocol-kit): zksync deployment address predict

### LINKED ISSUE #506: Predicted Safe address does not match deployed Safe address in zkSync network
## Description

During a Safe creation on zkSync, the predicted Safe address is different from the actual resulting address.

Check: https://github.com/safe-global/safe-wallet-web/issues/2370

## Expected result

Both addresses are the same.

dasanra:
## What it solves
Resolves #506

## How this PR fixes it
zkSync Era uses a different way to calculate deployments with Create2. The main difference is that the specific zksync deployedBytecode from the `SafeProxy` or the old `GnosisSafeProxy` it's also needed.

For zkSync Era we will use their explicit util in order to get the correct result at the time of predicting the Safe address, but deployedBytecodes need to be added explicitly when having a new Safe contract release


packages/protocol-kit/src/contracts/utils.ts
```diff
@@ -180,9 +189,24 @@ export async function predictSafeAddress({
...
-  const constructorData = toBuffer(
-    ethAdapter.encodeParameters(['address'], [safeContract.getAddress()])
-  ).toString('hex')
+  const input = ethAdapter.encodeParameters(['address'], [safeContract.getAddress()])
+
+  const chainId = await ethAdapter.getChainId()
+  // zkSync Era counterfactual deployment is calculated differently
+  // https://era.zksync.io/docs/reference/architecture/differences-with-ethereum.html#create-create2
+  if ([280, 324].includes(chainId)) {
+    const bytecodeHash = zksync.utils.hashBytecode(
```
mmv08, line 198: Does the hash change over time? If not, I'd keep track of hashes for bundle size efficiency instead of saving the bytecode.

dasanra: As far as I can see it should always be the same for the same bytecode. This will add complexity at the time of adding new contract versions, as it won't just be to copy the bytecode, but converting it to the hash in an easy to store format, 3 steps.

I can document the process

mmv08: Only two more releases are planned for the Safe account contract (1.4.1 and 1.5.0 LTS), so it should be worth the effort. It's not like we're adding new proxies every week.

mmv08: And for 1.4.1/1.5.0 I think the proxy bytecode will stay the same because I don't remember changing anything in the proxy 🤔

dasanra: Added some comments to guide the next one in the process. What do you think?

packages/protocol-kit/src/contracts/utils.ts
```diff
@@ -22,6 +24,24 @@ import { ContractNetworkConfig, SafeAccountConfig, SafeDeploymentConfig } from '
...
   '0xb1073742015cbcf5a3a4d9d1ae33ecf619439710b89475f92e2abd2117e90f90'
 
+const ZKSYNC_MAINNET = 324
+const ZKSYNC_TESTNET = 280
+// For bundle size efficiency we store SafeProxy.sol/GnosisSafeProxy.sol zksync bytecode hash
+// in hex mode. To get the values below we need to:
+// 1. Compile Safe smart contracts for zksync
+// 2. Get `deployedBytecode` from SafeProxy.json/GnosisSafeProxy.json
+// 3. Use zksync utils SDK to get the bytecode hash
+//    const = bytecodeHash = zkSyncUtils.hashBytecode(${deployedBytecode})
```
mmv08, line 34:
```suggestion
//    const bytecodeHash = zkSyncUtils.hashBytecode(${deployedBytecode})
```

mmv08 (approved):
Looks boss!

Issue #462, opened 2023-06-16
[Tests] ProtocolKit unit tests are not run


germartinez:
## Context / issue

There are no scripts to execute these tests: https://github.com/safe-global/safe-core-sdk/tree/main/packages/protocol-kit/tests/unit
Only e2e tests are executed: https://github.com/safe-global/safe-core-sdk/blob/main/packages/protocol-kit/hardhat.config.ts#L46

PR #515, opened 2023-08-30
fix: Add script to execute the tests

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/462

## How this PR fixes it
We have a missing test script in the `protocol-kit` for the unit tests

### DIFF
--- a/packages/protocol-kit/tests/unit/eip-712.test.ts
+++ b/packages/protocol-kit/tests/unit/eip-712.test.ts
@@ -13,9 +13,9 @@ const safeTransactionData: SafeTransactionData = {
   value: '111',
   data: '0x222',
   operation: 333,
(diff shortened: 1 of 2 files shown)


packages/protocol-kit/tests/unit/eip-712.test.ts
```diff
@@ -13,9 +13,9 @@ const safeTransactionData: SafeTransactionData = {
   value: '111',
   data: '0x222',
   operation: 333,
```
dasanra, line 15: Operation is not having a valid value.

yagopv: fixed

dasanra (approved):

PR #519, opened 2023-08-31
feat(protocol-kit): add Scroll Sepolia Testnet chainId

nick8319:
Adds missing Scroll Sepolia Testnet chain to protocol-kit.
https://github.com/safe-global/safe-deployments/pull/283


dasanra:
Added in PR https://github.com/safe-global/safe-core-sdk/pull/576

Issue #516, opened 2023-08-30
Cleanup `ethers` full import from the kits


dasanra:
## Description

In some kits, `ethers` is being used as an util importing the full library. As we want to keep the codebase to the minimum either devs decide to use `ethers` or `web3.js` we should be careful of not importing ethers fully when is not needed.

## Additional context

https://github.com/safe-global/safe-core-sdk/blob/5ce80a25d3e9011e2ac1e8b502cf7e2dc6c58407/packages/onramp-kit/src/packs/monerium/SafeMoneriumClient.ts#L1
https://github.com/safe-global/safe-core-sdk/blob/5ce80a25d3e9011e2ac1e8b502cf7e2dc6c58407/packages/onramp-kit/src/packs/monerium/signatures.ts#L1
https://github.com/safe-global/safe-core-sdk/blob/5ce80a25d3e9011e2ac1e8b502cf7e2dc6c58407/packages/auth-kit/src/AuthKitBasePack.ts#L1

PR #523, opened 2023-09-04
refactor(auth-kit): Reduce bundle size

yagopv:
## What it solves
This PR reduces the bundle size of the `auth-kit` by refactoring several packages. #516 

The instantiation of the `api-kit` involves the use of an `EthAdapter`. This adapter is only useful to validate the safe address with or without the network prefix so it is not really necessary to use the whole adapter. We are removing this dependency and changing it for a `chainId` property. We will use the `protocol-kit` to call the required `validateEip3770Address()` function.

The summary of changes are:

- Remove `EthAdapter` dependency from the `SafeApiKit` instantiation and add a `chainId` one
- Make `txServiceUrl` optional in the `SafeApiKit` instantiation. Map by chainId to our prod endpoints. We can continue to use this property to provide custom URLs
- Adding a mapping between `chainId` and our Transaction Service deployments (Info extracted from config service)
- Remove `getTxServiceBaseUrl()` utility function as it only concatenates an `/api` string to the Transaction Service url and can be done with a simple string template. We are also improving the usage of custom endpoints by avoiding the current concatenation method (Received some feedback from the community about this)
- Adapt `auth-kit`, `onramp-kit` packages code and examples
- Improve examples to use local packages instead deployed to `npm` ones. This allows us to easily test the examples locally when multiple packages are being modified

#### api-kit (BREAKING) 
- `SafeApiKit` instantiation
```js
old: constructor({ txServiceUrl, ethAdapter }: SafeApiKitConfig)
new: constructor({ chainId, txServiceUrl? }: SafeApiKitConfig)
```
- '/api' is not added to the txServiceUrl prop

#### auth-kit (MINOR) 
- Type changes
```js
old: abstract getProvider(): ethers.providers.ExternalProvider | null
new: abstract getProvider(): AuthKitEthereumProvider | null
```
- `getSafes()` add parameter
```js
old: async getSafes(txServiceUrl: string): Promise<string[]>
new: async getSafes(chainId: number, txServiceUrl?: string): Promise<string[]>
```
- `WebAuthModalPack` constructor config optional
```js
old: constructor(config: Web3AuthConfig)
new: constructor(config?: Web3AuthConfig)
```
- Type changes
```js
old: getProvider(): ExternalProvider | null
new: getProvider(): AuthKitEthereumProvider | null
---
old: export type Web3AuthEventListener = (...args: any[]) => void
new: export type Web3AuthEventListener = (...args: unknown[]) => void
```

### DIFF
--- a/packages/api-kit/src/SafeApiKit.ts
+++ b/packages/api-kit/src/SafeApiKit.ts
@@ -51,7 +51,7 @@ class SafeApiKit {
       const url = TRANSACTION_SERVICE_URLS[chainId]
       if (!url) {
         throw new TypeError(
-          `There is no transaction service deployment available for the chainId: ${chainId}. Please use the parameter txServiceUrl for a custom deployment`
+          `There is no transaction service available for chainId ${chainId}. Please set the txServiceUrl property to use a custom transaction service.`
         )
       }
 


packages/api-kit/README.md
```diff
@@ -47,25 +47,15 @@ npm run test
...
 ### Initialize the SafeApiKit
 
 ```js
 import SafeApiKit from '@safe-global/api-kit'
 
 const safeService = new SafeApiKit({
-  txServiceUrl: 'https://safe-transaction-mainnet.safe.global',
-  ethAdapter
+  chainId: 1,
+  // Optional. txServiceUrl should be used if no Safe deployment for the chainId
```
dasanra, line 57:
```suggestion
  // Optional. txServiceUrl must be used to set a custom service. For example on chains where Safe doesn't run services.
```

guides/integrating-the-safe-core-sdk.md
```diff
@@ -42,8 +42,16 @@ As stated in the introduction, the [Safe API Kit](https://github.com/safe-global
 ```js
 import SafeApiKit from '@safe-global/api-kit'
 
-const txServiceUrl = 'https://safe-transaction-mainnet.safe.global'
-const safeService = new SafeApiKit({ txServiceUrl, ethAdapter })
+const safeService = new SafeApiKit({ chainId })
+```
+
+Using the `chainId`, we will map internally the URL to the corresponding Safe Transaction Service endpoint. If we don't have a deployment for the desired chain, then the `txServiceUrl` should be used.
```
dasanra, line 48:
```suggestion
Using the `chainId` is enough for chains where Safe runs a Transaction Service. For those chains where Safe doesn't run a service, use the `txServiceUrl` parameter to set the custom service endpoint.
```

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -22,28 +22,45 @@ import {
...
+  constructor({ chainId, txServiceUrl }: SafeApiKitConfig) {
+    this.#chainId = chainId
+
+    if (txServiceUrl) {
+      this.#txServiceBaseUrl = txServiceUrl
+    } else {
+      const url = TRANSACTION_SERVICE_URLS[chainId]
+      if (!url) {
+        throw new TypeError(
+          `There is no transaction service deployment available for the chainId: ${chainId}. Please use the parameter txServiceUrl for a custom deployment`
```
dasanra, line 54:
```suggestion
          `There is no transaction service available for chainId ${chainId}. Please set the txServiceUrl property to use a custom transaction service.`
```

dasanra (approved):

Issue #14, opened 2021-03-23
Follow EIP 1271 for contract signatures


germartinez:
Follow EIP 1271 for contract signatures.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md

mikheevm:
sandbox for signature validation: https://codesandbox.io/s/blissful-dewdney-sc25n?file=/src/is-valid-signature.ts

signature validation logic is taken from https://github.com/authereum/is-valid-signature

yagopv:
As far as I understand we need to implement `isValidSignature` in some place in the repo

I implemented this in the [Monerium pack](https://github.com/safe-global/safe-core-sdk/blob/fc07ad724920b2a93602f86d61a994cc2e8d56b2/packages/onramp-kit/src/packs/monerium/SafeMoneriumClient.ts#L219-L262) indeed

yagopv:
We created new tickets to handle the support of the EIP-1271 so going to close this one

PR #529, opened 2023-09-12
feat(protocol-kit): EIP-1271 integration

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/14

## How this PR fixes it
We are adding the execution of the `isValidSignature(hash, signatures)` to the `Safe` class

More info:
[What can we do with this new method](https://ethereum.org/es/developers/tutorials/eip-1271-smart-contract-signatures)
[Safe implementation](https://github.com/safe-global/safe-contracts/blob/main/contracts/handler/CompatibilityFallbackHandler.sol)
[Safe docs](https://docs.safe.global/safe-smart-account/signatures/eip-1271)
[Safe docs](https://github.com/safe-global/safe-docs/pull/235/files#diff-36d179fbeb9dbe7b2729d66ba7e9a4a1e4162ab741a3d009397a520a8789b5bc)


packages/protocol-kit/src/managers/signatureManager.ts
```diff
@@ -0,0 +1,127 @@
...
+    safeContract?: SafeContract,
+    contractNetworks?: ContractNetworksConfig
+  ) {
+    this.#ethAdapter = ethAdapter
+    this.#safeContract = safeContract
+    this.#contractNetworks = contractNetworks
+  }
+
+  async init() {
+    const safeVersion = (await this.#safeContract?.getVersion()) ?? '1.3.0'
```
dasanra, line 31: There is a `DEFAULT_SAFE_VERSION` property that could be used here instead

packages/protocol-kit/src/Safe.ts
```diff
@@ -3,10 +3,11 @@ import {
   EthAdapter,
   OperationType,
   SafeMultisigTransactionResponse,
+  SafeMultisigConfirmationResponse,
   SafeSignature,
   SafeTransaction,
   SafeTransactionDataPartial,
-  SafeTransactionEIP712Args,
```
jamiegreen7, line 9: Start

jamiegreen7: 0x072088215Cb7395162C576904A4566E76e9F0Ff3

jamiegreen7 (approved):

PR #556, opened 2023-10-20
feat(protocol-kit): EIP-1271 integration

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/14

## ⚠️ Breaking Changes ⚠️

**protocol-kit**
- Rename `signTransactionHash()` to `signHash()` as the parameter is a hash that is not exclusive for transactions

**auth-kit**
- Expose an `AuthKitEthereumProvider` instead the `Eip1193Provider` coming from ethers


packages/protocol-kit/src/adapters/web3/Web3Adapter.ts
```diff
@@ -273,13 +273,13 @@ class Web3Adapter implements EthAdapter {
...
     methodVersion?: 'v3' | 'v4'
   ): Promise<string> {
     if (!this.#signerAddress) {
       throw new Error('This method requires a signer')
     }
-    const typedData = generateTypedData(safeTransactionEIP712Args)
+    const typedData = generateTypedData(safeEIP712Args)
     let method = 'eth_signTypedData_v3'
     if (methodVersion === 'v4') {
       method = 'eth_signTypedData_v4'
```
dasanra, line 285: Wouldn't be good to use the SigningMethod enum here too?

There are many places in the code where these strings are written directly. Now you created the enum maybe you should use it in all those places too

packages/protocol-kit/src/Safe.ts
```diff
@@ -562,21 +680,47 @@ class Safe {
...
-    if (signingMethod === 'eth_signTypedData_v4') {
+
+    if (method === 'eth_signTypedData_v4') {
       signature = await this.signTypedData(transaction, 'v4')
-    } else if (signingMethod === 'eth_signTypedData_v3') {
+    } else if (method === 'eth_signTypedData_v3') {
       signature = await this.signTypedData(transaction, 'v3')
-    } else if (signingMethod === 'eth_signTypedData') {
-      signature = await this.signTypedData(transaction)
+    } else if (method === 'eth_signTypedData') {
```
dasanra, line 691: Another example of a comparison as string of values that exist on SigningMethod enum, that is the default value used in the function header.

packages/protocol-kit/src/adapters/web3/Web3Adapter.ts
yagopv: yes totally, i created it at a later stage and didn't make the susbstitutions

packages/protocol-kit/src/adapters/web3/Web3Adapter.ts
```diff
@@ -32,6 +32,7 @@ import {
   getSignMessageLibContractInstance,
   getSimulateTxAccessorContractInstance
 } from './contracts/contractInstancesWeb3'
+import { SigningMethod } from '../..'
```
dasanra, line 35: Could we set a route here that is easier to understand?

`@safe-global/protocol-kit` or `@safe-global/protocol-kit/types`

packages/protocol-kit/src/utils/signatures/utils.ts
```diff
@@ -1,11 +1,16 @@
...
-  SafeTransactionEIP712Args
+  SafeEIP712Args,
+  SafeTransactionData
 } from '@safe-global/safe-core-sdk-types'
 import { bufferToHex, ecrecover, pubToAddress } from 'ethereumjs-util'
+import semverSatisfies from 'semver/functions/satisfies'
 import { sameString } from '../address'
 import { EthSafeSignature } from './SafeSignature'
+import { getEip712MessageTypes, getEip712TxTypes } from '../eip-712'
+import { SigningMethod } from '../..'
```
dasanra, line 13: Could we set a route here that is easier to understand?

`@safe-global/protocol-kit` or `@safe-global/protocol-kit/types`

packages/protocol-kit/tests/e2e/eip1271.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+        const validatedResponse2 = await safeSdk1.isValidSignature(messageHash, '0x')
+        chai.expect(validatedResponse2).to.be.true
+      }
+    )
+
+    itif(safeVersionDeployed >= '1.3.0')('should revert when message is not signed', async () => {
+      const { safeSdk1 } = await setupTests()
+
+      const response = await safeSdk1.isValidSignature(await hashSafeMessage(MESSAGE), '0x')
```
dasanra, line 153: Hash safeMessage not async

```suggestion
      const response = await safeSdk1.isValidSignature(hashSafeMessage(MESSAGE), '0x')
```

packages/protocol-kit/tests/e2e/eip1271.test.ts
```diff
@@ -0,0 +1,475 @@
...
+  describe('In the context of a 2/3 Safe and a 1/1 signer Safe account', async () => {
+    const setupTests = deployments.createFixture(async ({ deployments, getChainId }) => {
+      await deployments.fixture()
+      const accounts = await getAccounts()
+      const chainId = await getChainId()
+      const contractNetworks = await getContractNetworks(BigInt(chainId))
+      const fallbackHandlerAddress = contractNetworks[chainId].fallbackHandlerAddress
+      const [account1, account2] = accounts
+
+      // Create a 1/1 Safe to sign the messages
```
dasanra, line 52: Technically this would be a 1/2 right?
```suggestion
      // Create a 1/2 Safe to sign the messages
```

dasanra (approved):

Issue #543, opened 2023-10-05
[EIP1271] `api-kit` support


yagopv:
## Context / Issue

We should add support in the `api-kit` for the messages endpoints

## Proposed solution

Add endpoint support in the api-kit for the new endpoints:

[Docs](https://github.com/safe-global/safe-docs/pull/235/files#diff-36d179fbeb9dbe7b2729d66ba7e9a4a1e4162ab741a3d009397a520a8789b5bc)
[Swagger](https://safe-transaction-mainnet.safe.global/)

Consider a refactor and check for other missing endpoints. The kit needs some maintenance and separation of concerns (Currently all the endpoints under the same file

PR #557, opened 2023-10-20
feat(api-kit): Add messages endpoints

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/543

## How this PR fixes it
By adding the messages endpoints to the api-kit and corresponding e2e tests (Verify signatures using `protocol-kit` are  working)

```
GET /v1/messages/${messageHash}/
POST /v1/messages/${messageHash}/signatures/
GET /v1/safes/${safeAddress}/messages/
POST /v1/safes/${safeAddress}/messages/
```


packages/protocol-kit/src/Safe.ts
```diff
@@ -500,7 +500,7 @@ class Safe {
     }
 
     if (typeof txOrMessage === 'string') {
-      return soliditySha3(utf8ToHex(txOrMessage)) || ''
+      return ethers.hashMessage(txOrMessage) || ''
```
dasanra, line 503: `hashMessage` always return a string, the fallback shouldn't be necessary

dasanra (approved):

PR #559, opened 2023-10-22
fix: options added in executeRelayTransaction

donoso-eth:
## What it solves
**spondoredCalls are not routed properly**

In the account-abstraction-kit-poc@1.3.0 when calling
```
    const response = await this.#relayPack.executeRelayTransaction(
      signedSafeTransaction,
      this.#safeSdk
    )
```
the options are not passed so that the relay-kit will route all calls through `this.sendSyncTransaction(target, encodedTransaction, chainId, options)`

## How this PR fixes it
Adding the options to the executeRelayTransaction call
```
    const response = await this.#relayPack.executeRelayTransaction(
      signedSafeTransaction,
      this.#safeSdk,
      options
    )
```

### DIFF
--- a/packages/account-abstraction-kit/src/AccountAbstraction.ts
+++ b/packages/account-abstraction-kit/src/AccountAbstraction.ts
@@ -106,7 +106,8 @@ class AccountAbstraction {
 
     const response = await this.#relayPack.executeRelayTransaction(
       signedSafeTransaction,
-      this.#safeSdk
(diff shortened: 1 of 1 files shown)


donoso-eth:

dasanra:
@donoso-eth thank you for your PR! We fixed it in https://github.com/safe-global/safe-core-sdk/pull/571 adding some test cases to the package.

PR #560, opened 2023-10-23
Chain-specific Safe addresses

### LINKED ISSUE #464: Force the deployment of different Safe addresses in different networks
## Context / issue

The **default behaviour** of the Protocol Kit goes against the vision Safe has on having one address vs. multiple addresses when deploying a Safe to different networks.

## Proposed solution

The Protocol Kit is using the [createProxyWithNonce](https://github.com/safe-global/safe-contracts/blob/main/contracts/proxies/SafeProxyFactory.sol#L52) method with a default `saltNonce` to deploy Safes.

It looks like [createChainSpecificProxyWithNonce](https://github.com/safe-global/safe-contracts/blob/main/contracts/proxies/SafeProxyFactory.sol#LL67C37-L67C37) method does the same by calculating the `chainId` in the contract, but this makes it imposible to redeploy the same address in a different network if some funds need to be recovered. If some funds need to be recovered in another network we should keep it flexible and allow to deploy a Safe in the same address by adding the required `chainId` manually (different than the chain where the funds are stuck and the same as the chain of the original Safe).

For that reason we should add the `chainId` to the `saltNonce` in the SDK when passed to `createProxyWithNonce`.

## Expected result

- Even if the same configuration (owner list, threshold, fallback handler, etc.) is used to deploy a Safe in different networks, the Protocol Kit should generate different addresses.

- Nothing is broken for existing Safes that are already deployed when calculating the Safe address. Even if the `chainId` is added to the `saltNonce` we need to take into consideration that already deployed Safes were deployed with a saltNonce withouth the `chainId`.

tmjssz:
## What it solves
Resolves #464 

## How this PR fixes it

Adding the `getChainSpecificDefaultSaltNonce` function which appends the given `chainId` to the `PREDETERMINED_SALT_NONCE`. This results in different default addresses for the Safe contract on different chains.

### Example

Deployed two Safes with identical parameters on two different chains without defining a custom salt nonce. Received different addresses:
* Görli: [`0x6a724ffd8bab7ce14144cbfafa44994121ec8b16`](https://goerli.etherscan.io/address/0x6a724ffd8bab7ce14144cbfafa44994121ec8b16) 
* Mumbai: [`0x0c78086f0b398d044e54dcfdf86387942ef05713`](https://mumbai.polygonscan.com/address/0x0c78086f0b398d044e54dcfdf86387942ef05713)


tmjssz:

packages/protocol-kit/tests/e2e/createSafeDeploymentTransaction.test.ts
```diff
@@ -1,5 +1,6 @@
 import chai from 'chai'
 import chaiAsPromised from 'chai-as-promised'
+import { keccak256 } from 'ethereumjs-util'
```
dasanra, line 3: @DaniSomoza this lib is deprecated at your PR, could you recommend an alternative here?

packages/protocol-kit/src/contracts/utils.ts
```diff
@@ -152,6 +152,14 @@ const memoizedGetProxyCreationCode = createMemoizedFunction(
   }
 )
 
+/**
+ * Provides a default salt nonce with appended chainId to generate different addresses for the
+ * same Safe in different networks.
+ */
+export function getChainSpecificDefaultSaltNonce(chainId: number): string {
```
DaniSomoza, line 159: What is your opinion about to provide a full JSDoc for this function?. Here's a suggested format:

```javascript
/**
 * Provides a chain-specific default salt nonce for generating unique addresses 
 * for the same Safe configuration across different chains.
 * 
 * @param {number} chainId - The chain ID associated with the chain.
 * @returns {string} The chain-specific salt nonce in hexadecimal format.
 */
```

packages/protocol-kit/tests/e2e/createSafeDeploymentTransaction.test.ts
mmv08: https://github.com/paulmillr/noble-hashes

packages/protocol-kit/src/contracts/utils.ts
tmjssz: Yes makes total sense! Added it

packages/protocol-kit/tests/e2e/createSafeDeploymentTransaction.test.ts
tmjssz: Thx, I updated the PR to use `@noble/hashes` now.

packages/protocol-kit/src/contracts/utils.ts
mmv08: 🤓 : the largest number in javascript is (2^53) – 1, but chain IDs in ethereum theoretically can be 256 bits numbers

mmv08: From [here](https://eips.ethereum.org/EIPS/eip-1344): 
> Adds a new opcode CHAINID at 0x46, which uses 0 stack arguments. It pushes the current chain ID onto the stack. Chain ID is a 256-bit value. The operation costs G_base to execute.

tmjssz: Ah ok, so `BigInt` should be used? But would it not be better then to change it everywhere in the repo, instead of only in this function? Because the `ethAdapter.getChainId()` returns a simple `number` value and that is where the chainId comes from when being passed to this function.

mmv08: Ah, then it is irrelevant for this PR. Thanks for the explanation

DaniSomoza: Makes sense! I will update the type to BigInt in the PR of the Ethers v6 migration.

mmv08: BigInt or string both should work. Ethers also has a `BigNumberish` type

dasanra: I think that now we have BigInt available in Javascript is the way to go. Also most libraries I expect they will use BigInt as for example Ethers v6 already does.

We created a separate ticket to handle this as there are quite some changes needed. Thank you for your input @mmv08 

https://github.com/safe-global/safe-core-sdk/issues/564

dasanra (approved):

PR #585, opened 2023-11-15
chore: cleanup unused dependencies

dasanra:
## What it solves
 - Removes `ethers` as explicit dependency in those packages where is not used
 - Adds Safe v1.2.0 contracts directly in the project to avoid a HUGE amount of outdated dependencies.

## How this PR fixes it
Setting v1.2.0 contracts directly in the repo instead fetching them from npm, saves A LOT of space and removes a big chunk of unused and outdated dependencies. There were some `devDependencies` set as `dependencies` in safe-contracts v1.2.0 `package.json` making our `node_modules` folder to duplicate in size.
With this change we reduce `node_modules` size from 1.1GB to 665MB
We will also get rid of some outdated dependencies and it's associated alerts.


DaniSomoza (approved):
Approved! :3

PR #590, opened 2023-11-17
fix: createTransaction to accept only transactions array

### LINKED ISSUE #334: Non-editable options for proposals.
Hi!
I need to create some proposals and so i need to specify options with the correct "nonce", but these options are skipped here.
[https://github.com/safe-global/safe-core-sdk/blob/53a21d693467a3c1e47fe30fa4db28b4374af1b5/packages/safe-core-sdk/src/Safe.ts#L345-L347](https://github.com/safe-global/safe-core-sdk/blob/53a21d693467a3c1e47fe30fa4db28b4374af1b5/packages/safe-core-sdk/src/Safe.ts#L345-L347)
My goal is to create proposals, which will be approved one by one. 
And one more important point, txs shouldn't be packaged in MultiTx, bcs each of them can exceed the gas limit.

Maybe this code is more correct?
`newTransaction = isMetaTransactionArray(safeTransactionData)`  
 `       ? { ...options, ...safeTransactionData[0] }`
`       : { ...options, ...safeTransactionData }`

dasanra:
## What it solves
Resolves #334 

Create transaction accepted either an object or an array to create a new Safe transaction. This felxibility was OK, but it generates confusion and all scenarios can be covered by accepting an array only, as we also had an options object in place. Also this way, the protocol-kit create transaction aligns with the relay-kit method signature.

## How this PR fixes it
Accepts only an array of transactions, even if it only contains one transaction.

### DIFF
--- a/packages/protocol-kit/src/types/index.ts
+++ b/packages/protocol-kit/src/types/index.ts
@@ -128,7 +128,7 @@ export type ConnectSafeConfig =
   | ConnectSafeConfigWithPredictedSafe
 
 export interface CreateTransactionProps {
-  /** safeTransactionData - The transaction array to process */
+  /** transactions - The transaction array to process */
   transactions: MetaTransactionData[]
   /** options - The transaction array optional properties */
   options?: SafeTransactionOptionalProps
--- a/packages/protocol-kit/src/types/index.ts
+++ b/packages/protocol-kit/src/types/index.ts
@@ -128,8 +128,8 @@ export type ConnectSafeConfig =
   | ConnectSafeConfigWithPredictedSafe
 
 export interface CreateTransactionProps {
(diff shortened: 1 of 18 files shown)


packages/protocol-kit/src/types/index.ts
```diff
@@ -128,8 +128,8 @@ export type ConnectSafeConfig =
   | ConnectSafeConfigWithPredictedSafe
 
 export interface CreateTransactionProps {
-  /** safeTransactionData - The transaction or transaction array to process */
-  safeTransactionData: SafeTransactionDataPartial | MetaTransactionData[]
+  /** safeTransactionData - The transaction array to process */
```
tmjssz, line 131:
```suggestion
  /** transactions - The transaction array to process */
```

tmjssz (approved):

PR #599, opened 2023-11-22
fix exception handling in signature decoding for Monerium Pack

DaniSomoza:
## What it solves

Improve exception handling in signature decoding for Monerium.pack

Added a couple of unit tests

## How this PR fixes it

Now we are using `Promise.allSettled` instead of `Promise.all`
Added `parseIsValidSignatureErrorResponse` and `decodeSignatureData` to parse the error


dasanra (approved):
@yagopv you should also review this piece of code when you have a moment

PR #642, opened 2023-12-15
fix(onramp-kit): monerium/sdk now accepts chainId

einaralex:
## What it solves

v2.9.0 of monerium/sdk now accepts chainId


dasanra (approved):

dasanra:
Thank you for considering my suggestion so fast!

PR #655, opened 2024-01-09
feat: add short names for kroma networks

ElvisKrop:
## What it solves
Adds short names for both Kroma Mainnet and Kroma Sepolia Testnet:

- https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-255.json
- https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-2358.json

Blockers
**UPD:** NO BLOCKERS :tada: 

- ~~1.30.0 release of safe-deployments~~


dasanra:
Solved in #660

PR #666, opened 2024-01-16
feat: add short name for Mode Mainnet

ElvisKrop:
## What it solves
Adds short name for Mode Mainnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-34443.json

Contracts we merged into safe-deployment at https://github.com/safe-global/safe-deployments/pull/356

Thanks in advance!


dasanra:
Will be solved in https://github.com/safe-global/safe-core-sdk/pull/675

PR #676, opened 2024-01-24
feat(protocol-kit): Add AbiType for SignMessageLib 

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/653

## How this PR fixes it
Migrating SignMessageLib from Typechain to Abitype


yagopv:
We need to align around the `mapToTypechainContract` methods as if we want to remove them as seems to be the initial intention (and probably the right one) we are going to introduce a lot of breaking changes. 

If we keep them then we need to map to the old interfaces as in this case is the `SignMessageLibContract` in the `safe-core-sdk-types`

@dasanra @DaniSomoza @tmjssz

packages/protocol-kit/src/adapters/web3/contracts/SignMessageLib/v1.4.1/SignMessageLibContract_V1_4_1_Web3.ts
```diff
@@ -1,10 +1,119 @@
...
+      .estimateGas(options)
+      .then(BigInt)
+  }
+
+  getAddress: GetAddressSignMessageLibFunction = () => {
+    return Promise.resolve(this.contract.options.address)
+  }
+
+  async getMessageHash(args: readonly [string]): Promise<readonly [string]> {
+    return [await this.contract.methods.getMessageHash(...args)]
```
tmjssz, line 81: Should this not be here and also for the 1.3.0 implementation?
```suggestion
    return [await this.contract.methods.getMessageHash(...args).call()]
```

yagopv: true!

tmjssz (approved):

PR #691, opened 2024-02-07
chore: bump monerium sdk to 2.11.0 for Sepolia support

einaralex:
DO NOT MERGE
Wasn't able to properly test since the onramp kit seems to still trying to use Goerli.
Tests are failing, because they are trying to use Goerli.

## What it solves

Version 2.11.0 of `@monerium/sdk` supports Sepolia.


dasanra:
@einaralex we will take care to bump to the latest version in this PR https://github.com/safe-global/safe-core-sdk/pull/788

Thank you.

PR #703, opened 2024-02-19
feat: add short name for Sei Devnet

ElvisKrop:
## What it solves
Adds short name for Sei Devnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-713715.json


dasanra:

PR #706, opened 2024-02-22
feat: add short name for Lisk Sepolia Testnet

ElvisKrop:
## What it solves
Adds short name for Lisk Sepolia Testnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-4202.json


dasanra:
Will be closed in https://github.com/safe-global/safe-core-sdk/pull/752

Issue #699, opened 2024-02-15
[4337] Generate init code and deployment flow


dasanra:
## Context / issue

It should be possible to deploy a 1/N Safe with the module enabled.

## Proposed solution

Generate the init code. Create the deployment transaction. If possible should use a batch to be able to add other transactions if necessary.

PR #707, opened 2024-02-22
feat: Initialize new safe (`initCode`)

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/699

## How this PR fixes it
This PR add support to deploy a Safe (counterfactual) by using the `initCode` field of the `UserOperation` sent to the bundler


packages/relay-kit/package.json
```diff
@@ -38,8 +38,10 @@
     "@gelatonetwork/relay-sdk": "^5.5.0",
     "@safe-global/protocol-kit": "^3.0.0",
     "@safe-global/safe-core-sdk-types": "^4.0.0",
+    "@safe-global/safe-modules-deployments": "^2.0.0",
+    "alchemy-sdk": "^3.1.2",
+    "ethers": "^6.7.1",
     "permissionless": "^0.0.35",
```
dasanra, line 44: This shouldn't be necessary any more right?

packages/relay-kit/src/packs/safe-4337/types.ts
```diff
@@ -1,19 +1,41 @@
...
-  readonly signatures: Map<string, SafeSignature>
-  getSignature(signer: string): SafeSignature | undefined
-  addSignature(signature: SafeSignature): void
-  encodedSignatures(): string
+export type Safe4337Options = {
+  protocolKit: Safe
+  bundlerUrl: string
+  rpcUrl: string
+  bundlerClient: ethers.JsonRpcProvider
+  publicClient: ethers.JsonRpcProvider
```
dasanra, line 35: This will make the kit only compatible with Ethers?

yagopv: Internally, yes. Currently, we use ethers throughout the entire SDK. In the future, we may switch to viem once (if :)) we eliminate the dependency.

The `EthersAdapter` is the only compatible option because web3 [lacks](https://github.com/safe-global/safe-core-sdk/blob/d4923af26fa58499ec112039300ad381e61a7542/packages/protocol-kit/src/adapters/web3/Web3Adapter.ts#L276-L319) a `signTypedData` function. We cannot depend on the EthAdapter interface since is doing specific SafeTx or SafeMessage things (not SafeOp). Without adding this logic in the protocol-kit, those functions cannot be used. The current logic resides in the pack

We can refactor the protocol-kit to expose the concrete signTypedData functions instead of the "safe" signTypedData ones. I probably can make it compatible with Web3Adapter by duplicating from the protocol-kit the  necessary logic in the pack and checking the type of the instance passed as param. Currently, it only works with EthersAdapter.

packages/relay-kit/package.json
yagopv: Yes, I removed permissionless and viem

We need to discuss if we want to provide the content of the providers folder as part of the sdk or as a guide in the examples. If we remove this gas calculations then i can remove the alchemy sdk as well

packages/relay-kit/package.json
```diff
@@ -38,8 +38,8 @@
     "@gelatonetwork/relay-sdk": "^5.5.0",
     "@safe-global/protocol-kit": "^3.0.0",
     "@safe-global/safe-core-sdk-types": "^4.0.0",
-    "permissionless": "^0.0.35",
-    "viem": "^2.7.6",
+    "@safe-global/safe-modules-deployments": "^2.0.0",
+    "alchemy-sdk": "^3.1.2",
```
yagopv, line 42: I would remove this and the AlchemyFeeProvider from the SDK

packages/relay-kit/src/packs/gelato/GelatoRelayPack.ts
```diff
@@ -50,14 +61,45 @@ export class GelatoRelayPack extends RelayKitBasePack {
...
+    gasToken?: string
+  ): Promise<GelatoEstimateFeeOptionsResult | string> {
+    let estimateFeeChainId: bigint
+    let estimateFeeGasLimit: bigint | undefined
+    let estimateFeeGasToken: string | undefined
+
+    if (typeof optionsOrChainId === 'object') {
+      estimateFeeChainId = optionsOrChainId.chainId
+      estimateFeeGasLimit = BigInt(optionsOrChainId.gasLimit)
+      estimateFeeGasToken = optionsOrChainId.gasToken || ZERO_ADDRESS
```
dasanra, line 88: We weren't adding the fallback `ZERO_ADDRESS` value before

dasanra (approved):

dasanra (approved):

PR #711, opened 2024-02-26
feat(protocol-kit): add short name for Zora Testnet

evseevnn:
## What it solves
Adds short names for Zora Testnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-999999999.json


dasanra:

PR #712, opened 2024-02-26
feat(protocol-kit): add short name for Mantle Sepolia Testnet

evseevnn:
## What it solves

Adds short names for Mantle Sepolia Testnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-5003.json


dasanra:

PR #714, opened 2024-02-26
feat(protocol-kit): add short name for Blast Sepolia Testnet

evseevnn:
## What it solves

Adds short names for Blast Sepolia Testnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-168587773.json


dasanra:
Solved in https://github.com/safe-global/safe-core-sdk/pull/752

PR #715, opened 2024-02-26
feat(protocol-kit): add short name for IoTeX Network Testnet

evseevnn:
## What it solves
Adds short name for IoTeX Network Testnet:

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-4690.json


dasanra:
To add shortnames the chain assets should be available at [`safe-deployments`](https://github.com/safe-global/safe-deployments) first.

evseevnn:
https://github.com/safe-global/safe-deployments/pull/421

PR #760, opened 2024-04-05
Remove Typechain from Safe contracts

### LINKED ISSUE #728: Remove Typechain from Safe contracts
## Context / issue

The current typing library is deprecated and this prevent us to update to Web3.js v4 or adding Viem as an adapter in the `protocol-kit`. Also it's way more complex than current solutions adding more complexity to the SDK maintenance.

https://github.com/safe-global/safe-core-sdk/issues/577

## Proposed solution

The Safe contracts already have AbiType typings created, but we are currently mapping them to Typechain compatible types. Many of the places are marked as TODO remove typechain. Some of the pending actions are:

 - Remove `mapToTypechainContract` in all Safe types and use the AbiType as the typing system.
 - Add missing `getNonce` and `getVersion` function to keep the same API.
 - Move all possible methods to the base class when shared across all versions.
 - `setup` method is only used for testing, but added in the global types. Remove if possible.

DaniSomoza:
## What it solves
Resolves #728


packages/protocol-kit/src/types/index.ts
```diff
@@ -2,7 +2,7 @@ import { SafeTransactionOptionalProps } from '@safe-global/protocol-kit/utils/tr
 import {
   EthAdapter,
   MetaTransactionData,
-  SafeContract,
+  SafeContactType,
```
dasanra, line 5: There is a typo here:
```suggestion
  SafeContractType,
```

packages/protocol-kit/src/types/index.ts
```diff
@@ -159,7 +159,7 @@ export interface SwapOwnerTxParams {
 
 type StandardizeSafeTxDataWithSafeContractProps = {
   /** safeContract - The Safe contract to use */
-  safeContract: SafeContract
+  safeContract: SafeContactType
```
dasanra, line 162: There is a typo here. Review the code where this parameter is used
```suggestion
  safeContract: SafeContractType
```

packages/protocol-kit/src/types/abi.d.ts
```diff
@@ -1,18 +0,0 @@
...
-    //   inputs: `0x${string}` | Uint8Array
-    //   outputs: `0x${string}`
-    // }
-    AddressType: string
-    BytesType: {
-      inputs: string
-      outputs: string
-    }
-  }
-}
```
dasanra, line 18: Could you declare the module like this instead of adding it to `safe-core-sdk-types/src/index`

packages/protocol-kit/src/adapters/ethers/utils/index.ts
```diff
@@ -1,5 +1,5 @@
 import { ContractTransactionResponse, Provider, AbstractSigner } from 'ethers'
-import { EthersTransactionOptions, EthersTransactionResult } from '../types'
+import { EthersTransactionOptions, EthersTransactionResult } from 'packages/safe-core-sdk-types'
```
dasanra, line 3:
```suggestion
import { EthersTransactionOptions, EthersTransactionResult } from '@safe-global/safe-core-sdk-types'
```

Issue #651, opened 2024-01-04
[EIP-1193] Implement an EIP-1193 provider instead of adapters


mmv08:
## Context / issue

When working with the SDK, the adapter's concept adds some cognitive complexity (I haven't seen it in any other SDK). It is also another piece that has to be maintained.

## Proposed solution

There's a standard for Ethereum Provider JavaScript API: https://eips.ethereum.org/EIPS/eip-1193

All major libs are capable of exporting such a provider. Some of them have it built-in, and other ones require some custom code.

- Ethers: https://docs.ethers.org/v5/api/experimental/#experimental-eip1193bridge
- Viem
> You build it yourself. It's pretty much just const provider = { request: client.request } 😝
> Throw on events with the eventemitter3 lib if you need account/chain changed events
- Web3.js: https://docs.web3js.org/api/web3-utils/class/Eip1193Provider/

## Alternatives

- Keep maintaining adapters

## Additional context

I've used this approach in a different library, and it has worked well so far: https://github.com/gnosis/evm-proxy-detection

PR #770, opened 2024-04-10
feat(protocol-kit): Add EIP-1193 provider support

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/651

## How this PR fixes it
- Remove the `EthersAdapter` and `Web3Adapter`, and add the `SafeProvider` with standard EIP-1193 support and additional Safe-specific functionality. The `SafeProvider` is instantiated internally, so there is no need to add it as an input parameter as with the adapters.

`signer` is optional so if not specified will try to get the first connected one in the provider (`eth_accounts`, `eth_requestAccounts`)

```typescript
// before
await Safe.create({
   ethAdapter: new EthersAdapter({ ethers, signerOrProvider })
   safeAddress: '0xSafeAddress'
   ...
})

// now
await Safe.create({
   provider: window.ethereum, // Or any compatible EIP-1193 provider
   signer: 'signerAddressOrPrivateKey'
   safeAddress: '0xSafeAddress'
   ...
})

// ...or...
await Safe.create({
   provider: 'http://rpc.url', //Or websocket
   signer: 'signerAddressOrPrivateKey'
   safeAddress: '0xSafeAddress'
   ...
})
```

This change make the `AccountAbstraction` and `Safe4337Pack` instantiation different as well as the dependency with the `EthAdapter` is changed to provider + signer (Same as Safe.create)

- e2e tests
  - Remove ganache tests
  - Add `viem` support tests in e2e action and scripts (same as `ethers` and `web3`)
  - Refactor `api-kit` tests
  - Refactor `protocol-kit` tests
- Remove all Web3 specific dependencies (web3-utils, web3-core) and contracts
  -  Remove `gas` from `TransactionOptions` and specific return types from web3
  - Simplify types as the `Safe*ImplementationType`'s and `DeepWriteable`'s
 - Remove all `Ethers` suffixes as now are the unique contract types and files


packages/protocol-kit/tests/e2e/eip1271-contract-signatures.test.ts
```diff
@@ -77,14 +77,14 @@ describe('The EIP1271 implementation', () => {
...
-        const ethAdapter1 = await getEthAdapter(account1.signer)
-        const ethAdapter2 = await getEthAdapter(account2.signer)
-        const ethAdapter3 = await getEthAdapter(account3.signer)
-        const ethAdapter4 = await getEthAdapter(account4.signer)
-        const ethAdapter5 = await getEthAdapter(account5.signer)
+        const provider1 = getEip1193Provider()
+        const provider2 = getEip1193Provider()
+        const provider3 = getEip1193Provider()
+        const provider4 = getEip1193Provider()
+        const provider5 = getEip1193Provider()
```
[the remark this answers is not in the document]
yagopv: Sure, I was doing global replacements and removing the signer from the original `getEthAdapter`  was one of the latest things I did, so I haven't yet seen this kind of issues yet 🙃. These should be addressed once I review the entire structure and code

packages/api-kit/tests/utils/setupKits.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type GetKitsOptions = {
+  signer?: SafeProviderConfig['signer']
+  txServiceUrl?: string
+  safeAddress: string
+}
+
+function getEip1193Provider(): Eip1193Provider {
+  switch (process.env.ETH_LIB) {
```
dasanra, line 19: Could we add Viem here too?

packages/safe-core-sdk-types/src/types.ts
```diff
@@ -99,8 +97,7 @@ export interface BaseTransactionResult {
 }
 
 export interface TransactionResult extends BaseTransactionResult {
-  promiEvent?: PromiEvent<TransactionReceipt>
-  transactionResponse?: ContractTransactionResponse
+  transactionResponse: ContractTransactionResponse
```
dasanra, line 100: This could generate a breaking change later on if we go into Viem internally

packages/api-kit/tests/utils/setupKits.ts
yagopv: yes i can. Anyway in the github action only the ethers script it's being executed. Do you think is worth to add the matrix execution as in the protocol-kit?

yagopv: Ok i added viem and some more things as removing the hardhat web3 specific library, renaming the jobs both filenames and internally because it's never clear what each action does. I'm executing the e2e in the api-kit for the 3 libraries but couldn't do it in parallel (matrix) because some of them fail. In sequence seems to work

packages/safe-core-sdk-types/src/types.ts
yagopv: mmm i removed the promievent because is web3 specific. ContractTransactionResponse is ethers specific yes, do you mean we should use a internal type ?

.github/workflows/api-kit-e2e-test.yml
```diff
@@ -21,4 +21,7 @@ jobs:
       - run: |
           yarn install --frozen-lockfile
           yarn build
```
dasanra, line 23: Let's separate the steps so we split in shorter log chunks when we check the action.

```suggestion
      - name: Yarn install
        run: yarn install --frozen-lockfile

      - name: Build
        run: yarn build

      - name: Test
        run: |
```

packages/safe-core-sdk-types/src/types.ts
yagopv: Hey i did several things here:
- I realized the `EthersTransactionOptions` after removing the `web3` specific stuff is duplicated with the `TransactionOptions`. Same for `EthersTransactionResult` so I removed them and updated a lot of generic types. 
- I renamed the `AdapterSpecificContractFunction` to `SafeContractFunction`
- Removed the ethers dependency from `safe-core-sdk-types` as it's no longer needed
- I set the `TransactionResponse` as unkwown but not completely sure about if this is the best option. It works without more changes at the SDK level

cc: @DaniSomoza @tmjssz

guides/integrating-the-safe-core-sdk.md
```diff
@@ -24,16 +24,14 @@ To integrate the [Safe Core SDK](https://github.com/safe-global/safe-core-sdk) i
 
 ## <a name="initialize-sdks">2. Initialize the SDK’s</a>
 
-### Instantiate an EthAdapter
+### Select your Ethereum `provider` and `signer`
 
-First of all, we need to create an `EthAdapter`, which contains all the required utilities for the SDKs to interact with the blockchain. It acts as a wrapper for [web3.js](https://web3js.readthedocs.io/) or [ethers.js](https://docs.ethers.org/v6/) Ethereum libraries.
+To use our kits, you need to provide an Ethereum provider and a signer. The provider is the connection to the Ethereum network, while the signer is the account that will sign the transactions (The safe owner). When using an injected provider like MetaMask, the signer is the account selected in the wallet.
```
dasanra, line 29:
```suggestion
To use our kits, you need to provide an Ethereum provider and a signer. The provider is the connection to the Ethereum network, while the signer is an account that will sign the transactions (a Safe owner). When using an injected provider like MetaMask, the signer is the account selected in the wallet.
```

guides/integrating-the-safe-core-sdk.md
```diff
@@ -24,16 +24,14 @@ To integrate the [Safe Core SDK](https://github.com/safe-global/safe-core-sdk) i
...
 
-Depending on the library used by the Dapp, there are two options:
+In the examples below, you can see `provider` and `signer` properties, which represent:
 
-- [Create an `EthersAdapter` instance](https://github.com/safe-global/safe-core-sdk/tree/main/packages/protocol-kit/src/adapters/ethers)
-- [Create a `Web3Adapter` instance](https://github.com/safe-global/safe-core-sdk/tree/main/packages/protocol-kit/src/adapters/web3)
-
-Once the instance of `EthersAdapter` or `Web3Adapter` is created, it can be used in the SDK initialization.
+- `provider`: You can provide an EIP-1193 compatible provider or an HTTP/WebSocket RPC URL.
+- `signer`: This is optional. If not provided, we will try to get the connected account of the provider. If you provide it, it represents either the provider's address you want to use or a private key, which we will use internally to retrieve the provider's address.
```
dasanra, line 34:
```suggestion
- `signer`: This is an optional parameter. It should be the provider's address you want to use or a private key. If not set, it will try to fetch a connected account from the provider.
```

packages/api-kit/tests/e2e/getTransactionConfirmations.test.ts
```diff
@@ -21,13 +19,6 @@ describe('getTransactionConfirmations', () => {
       .to.be.rejectedWith('Invalid safeTxHash')
   })
 
-  it.skip('should return an empty array if the safeTxHash is not found', async () => {
```
dasanra, line 24: Why was this test removed?

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -251,39 +238,13 @@ describe('Off-chain signatures', () => {
       }
     )
 
-    itif(process.env.ETH_LIB === 'web3')(
```
dasanra, line 254: If we remove this test wouldn't make sense to remove the `itif` from the previous one?

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -300,37 +261,11 @@ describe('Off-chain signatures', () => {
...
```
dasanra, line 303: If we remove this test wouldn't make sense to remove the `itif` from the previous one?

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -226,111 +215,51 @@ describe('Off-chain signatures', () => {
...
-          to: safeAddress,
-          value: '0',
-          data: '0x'
-        }
-        const tx = await safeSdk.createTransaction({ transactions: [safeTransactionData] })
-        chai.expect(tx.signatures.size).to.be.eq(0)
-        const signedTx = await safeSdk.signTransaction(tx, SigningMethod.ETH_SIGN_TYPED_DATA_V3)
-        chai.expect(tx.signatures.size).to.be.eq(0)
-        chai.expect(signedTx.signatures.size).to.be.eq(1)
+    it('should add the signature of the current signer using eth_signTypedData with ethers provider', async () => {
```
dasanra, line 218:
```suggestion
    it('should add the signature of the current signer using eth_signTypedData', async () => {
```

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -226,111 +215,51 @@ describe('Off-chain signatures', () => {
...
-        const safeTransactionData = {
-          to: safeAddress,
-          value: '0',
-          data: '0x'
-        }
-        const tx = await safeSdk.createTransaction({ transactions: [safeTransactionData] })
-        await chai
-          .expect(safeSdk.signTransaction(tx, SigningMethod.ETH_SIGN_TYPED_DATA_V3))
-          .to.be.rejectedWith("EIP-712 is not supported by user's wallet")
+    it('should add the signature of the current signer using eth_signTypedData_v3 with ethers provider', async () => {
```
dasanra, line 238:
```suggestion
    it('should add the signature of the current signer using eth_signTypedData_v3', async () => {
```

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
yagopv: yes!

packages/api-kit/tests/e2e/getTransactionConfirmations.test.ts
yagopv: I didn't think much here and removed it because it was being skipped. It was likely skipped because it was throwing an error if you didn't set a correct hash, so I fixed it and changed the hash

dasanra (approved):

PR #777, opened 2024-04-19
feat(api-kit): 4337 API functions

### LINKED ISSUE #775: [4337] API Compatibility
## Context / issue

Add to the `api-kit` the necessary methods to be able to fetch information about SafeOperations

## Proposed solution

Add the following methods to the `api-kit`:
 - GET `/v1/safe-operations/{safe_operation_hash}/`
 - GET `/v1/safes/{address}/safe-operations/`
 - POST `/v1/safes/{address}/safe-operations/`
 
 Remember to add testing suite for the endpoints above.

## Other links

[Staging Sepolia swagger](https://safe-transaction-sepolia.staging.5afe.dev/)

tmjssz:
## What it solves
Resolves #775

Adding support for the following 4337 API endpoints:

* GET `/v1/safe-operations/{safe_operation_hash}/`
* GET `/v1/safes/{address}/safe-operations/`
* POST `/v1/safes/{address}/safe-operations/`


packages/api-kit/src/types/safeTransactionServiceTypes.ts
```diff
@@ -3,6 +3,7 @@ import {
   SafeMultisigTransactionResponse,
   SafeTransactionData
 } from '@safe-global/safe-core-sdk-types'
+import SafeOperation from '@safe-global/relay-kit/src/packs/safe-4337/SafeOperation'
```
dasanra, line 6: Here we are consumin a class from another kit that we are not importing in the dependencies. We should minimize cross-package dependencies so it's likely we need the type to be defined in `safe-core-sdk-types`

tmjssz: Updated accordingly and fixed tests as well. Ready to be reviewed again

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -715,6 +719,110 @@ class SafeApiKit {
...
+   * @throws "Safe address must not be empty"
+   * @throws "Invalid Safe address {safeAddress}"
+   * @throws "Module address must not be empty"
+   * @throws "Invalid module address {moduleAddress}"
+   * @throws "SafeOperation is not signed by the given signer {signerAddress}"
+   */
+  async addSafeOperation({
+    moduleAddress: moduleAddressProp,
+    safeAddress: safeAddressProp,
+    safeOperation,
```
leonardotc, line 772: So... I dont understand the use case for this particular function but here is some food for thought. Right now, there is no internal coupling between this and the Safe4337 module which is better then if we were importing the type from there. There is, however, an implicit one. For someone to use this endpoint, they would have to either: use `Safe4337Pack#createTransaction` or implementing an interface themselves that accomplishes the same thing (which is a bit awkward). Are there situation where all these things could be supplied by an external library or people should ALWAYS use both modules together? Should we really use such complex object just for an api proxy and should the api-kit itself sign in this case? Perhaps the interface here is just really an `UserOperation` (which is clearly defined by the ERC) and the fields associated with packing signature (`validUntil`, `validAfter` and the authorizer's `signature`). That data field also almost look like something to represent the internal state of the object.

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -715,6 +719,110 @@ class SafeApiKit {
...
+    }
+
+    try {
+      moduleAddress = this.#getEip3770Address(moduleAddressProp).address
+    } catch (err) {
+      throw new Error(`Invalid module address ${moduleAddressProp}`)
+    }
+
+    const signerAddress = await signer.getAddress()
+    const signature = safeOperation.getSignature(signerAddress)
```
leonardotc, line 797: I was looking at the server-side code and this field seems to be validated like this:
```
safe_owners = self._get_owners(safe_address)
parsed_signatures = SafeSignature.parse_signature(
    signature, safe_operation_hash, safe_operation_hash_preimage
)
owners_processed = set()
safe_signatures = []
for safe_signature in parsed_signatures:
    # do stuff
```
Please just check if those are not every owner's concatenated signature. I think the `SafeOperation#encodedSignatures()` or something.

packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -645,6 +645,77 @@ describe('Endpoint tests', () => {
...
+            safeAddress,
+            safeOperation,
+            signer
+          })
+        )
+        .to.be.eventually.deep.equals({ data: { success: true } })
+
+      // We need to get the object without the "callGasLimit" prop because the api endpoint expects it as "callDataGasLimit"
+      /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+      const { callGasLimit, ...safeOperationWithoutCallGasLimit } = safeOperation.data
```
leonardotc, line 706: I am not sure but I think if you do something like
```
const { callGasLimit: callDataGasLimit, ...safeOperationWithoutCallGasLimit } = safeOperation.data
```
you can use the variable below and remove the eslint comment.

packages/api-kit/tests/e2e/addSafeOperation.test.ts
```diff
@@ -0,0 +1,193 @@
...
+import { EthersAdapter } from 'packages/protocol-kit'
+import { getSafe4337ModuleDeployment } from '@safe-global/safe-modules-deployments'
+import config from '../utils/config'
+import { getServiceClient } from '../utils/setupServiceClient'
+
+chai.use(chaiAsPromised)
+chai.use(sinonChai)
+
+const SIGNER_PK = '0x83a415ca62e11f5fa5567e98450d0f82ae19ff36ef876c10a8d448c788a53676'
+const SAFE_ADDRESS = '0x60C4Ab82D06Fd7dFE9517e17736C2Dcc77443EF0' // 1/1 Safe (v1.4.1) with signer above as owner + 4337 module enabled
```
leonardotc, line 20: The comment is great! Its not big deal but fo future reference: whenever there is signature logic for a multisig, Id go with a higher threshold else the tests would not cover that logic that much.

packages/api-kit/tests/e2e/getSafeOperation.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    })
+
+    it('should fail if safeOperationHash is invalid', async () => {
+      await chai.expect(safeApiKit.getSafeOperation('0x123')).to.be.rejectedWith('Not found.')
+    })
+  })
+
+  it('should get the SafeOperation', async () => {
+    const safeOperations = await safeApiKit.getSafeOperationsByAddress(SAFE_ADDRESS)
+    const safeOperationHash = safeOperations.results[0].safeOperationHash
```
leonardotc, line 33: Just a sugestion: consider on asserting over the array's length before pulling data. In this case, if that is zero (for whatever reason), you would have got `typeError: Cannot read properties of undefined` or something like that while, with the assertion, you would get an assertion error (which is usually easier to debug).

packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -645,6 +645,77 @@ describe('Endpoint tests', () => {
...
+    })
+
+    it('addSafeOperation', async () => {
+      const signature = '0xsignature'
+      const moduleAddress = '0xa581c4A4DB7175302464fF3C06380BC3270b4037'
+
+      const safeOperation = {
+        data: {
+          nonce: 42,
+          initCode: '0x123initcode',
```
leonardotc, line 678: I know this is mocked but it might have been cool to use structurally valid data. I think in this case this would have failed with a bad request since their serializer expect to be a valid hex string.

packages/safe-core-sdk-types/src/types.ts
```diff
@@ -252,3 +252,51 @@ export interface MetaTransactionOptions {
...
+
+export type EstimateGasData = {
+  maxFeePerGas?: bigint
+  maxPriorityFeePerGas?: bigint
+  preVerificationGas?: bigint
+  verificationGasLimit?: bigint
+  callGasLimit?: bigint
+}
+
+export interface SafeOperation {
```
leonardotc, line 294: I know all things are package-scoped but I it would be cool if we could avoid having things with the same name. Some developers will do a deep search and find two things and will not be sure which one is which.

tmjssz: Good point! I just wonder which name would make sense then. Because the purpose of the interface here is to allow importing the type from `safe-core-sdk-types` instead of from the relay-kit, where it's implemented. 
For consistency, maybe we could rename the [SafeOperation class](https://github.com/safe-global/safe-core-sdk/blob/feat/4337-api-integration/packages/relay-kit/src/packs/safe-4337/SafeOperation.ts) to `EthSafeOperation` so it would be the same pattern as e.g. for [EthSafeTransaction](https://github.com/safe-global/safe-core-sdk/blob/feat/4337-api-integration/packages/protocol-kit/src/utils/transactions/SafeTransaction.ts).

packages/api-kit/tests/endpoint/index.test.ts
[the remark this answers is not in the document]
tmjssz: Done!

packages/api-kit/tests/endpoint/index.test.ts
tmjssz: Ahh yes of course, that is much better. Thx!

packages/api-kit/tests/e2e/getSafeOperation.test.ts
tmjssz: Done

packages/api-kit/src/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+import { EMPTY_DATA } from './constants'
+
+export const isEmptyHexData = (input: string) => !input || input === EMPTY_DATA
```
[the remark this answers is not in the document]
tmjssz: Renamed the function

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -715,6 +719,110 @@ class SafeApiKit {
...
   }
+
+  /**
+   * Get the SafeOperations that were sent from a particular address.
+   * @param safeAddress - The Safe address to retrieve SafeOperations for
+   * @throws "Safe address must not be empty"
+   * @throws "Invalid Ethereum address {safeAddress}"
+   * @returns The SafeOperations sent from the given Safe's address
+   */
+  async getSafeOperationsByAddress(safeAddress: string): Promise<GetSafeOperationListResponse> {
```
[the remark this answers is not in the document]
tmjssz: Was not on purpose 😅 added the missing params now

packages/api-kit/.env.example
```diff
@@ -1,2 +1,3 @@
 MNEMONIC=
-PK=
\ No newline at end of file
+PK=
+PIMLICO_API_KEY=
```
dasanra, line 3: This is not used, right?

```suggestion
```

tmjssz: Yes true, it's not used anymore

packages/safe-core-sdk-types/src/types.ts
yagopv: One typical pattern is add the "I" letter as prefix so SafeOperation implements ISafeOperation

tmjssz: Thought about this already, but not a big fan of it. Also found this in the [TS style guide](https://ts.dev/style/#naming-style):
> Do not mark interfaces specially (IMyInterface or MyFooInterface) unless it's idiomatic in its environment. When introducing an interface for a class, give it a name that expresses why the interface exists in the first place (e.g. class TodoItem and interface TodoItemStorage if the interface expresses the format used for storage/serialization in JSON).

yagopv: You can go with the EthSafeOperation then

yagopv (approved):

PR #785, opened 2024-04-23
Improve readme: Add Preperation section

Citrullin:
## What it solves
When a user freshly clones the repository and builds/tests one specific package, it will fail due to missing types.

## How this PR fixes it
Just adds a little section in the readme that notifies the reader over this behaviour.


Citrullin:
recheck

dasanra:
@Citrullin thank you for your suggestion. We will be covering this topic [adding a CONTRIBUTING file](https://github.com/safe-global/safe-core-sdk/pull/815) explaining a bit better how to use this repository. We would still keep on top the instructions for those wanting to use the different packages.

PR #787, opened 2024-04-26
fix(protocol kit): Get modules paginated incorrect interface

leonardotc:
## What it solves
The method `getModulesPaginated` should return a tuple containing both the modules as well as the next address of the page.

## How this PR fixes it
- It changes the interface by adding the missing property as a "complex type".
- It adds the type since that is the result of the interface
- It adds assertions to test that scenario

### DIFF
--- a/packages/protocol-kit/hardhat/deploy/deploy-contracts.ts
+++ b/packages/protocol-kit/hardhat/deploy/deploy-contracts.ts
@@ -155,6 +155,20 @@ const deploy: DeployFunction = async (hre: HardhatRuntimeEnvironment): Promise<v
     deterministicDeployment: true
   })
 
+  await deploy('StateChannelModule', {
+    from: deployer,
+    args: [],
+    log: true,
+    deterministicDeployment: true
+  })
+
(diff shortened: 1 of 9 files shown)


leonardotc:
This was added to the `1.4.1` of the contract (https://github.com/safe-global/safe-smart-account/blob/v1.4.1/contracts/base/ModuleManager.sol:
```
        /**
          Because of the argument validation, we can assume that the loop will always iterate over the valid module list values
          and the `next` variable will either be an enabled module or a sentinel address (signalling the end). 
          
          If we haven't reached the end inside the loop, we need to set the next pointer to the last element of the modules array
          because the `next` variable (which is a module by itself) acting as a pointer to the start of the next page is neither 
          included to the current page, nor will it be included in the next one if you pass it as a start.
        */
        if (next != SENTINEL_MODULES) {
            next = array[moduleCount - 1];
        }
```

yagopv (approved):

yagopv:
Beautiful @leonardotc thanks!. I think you need to pull from the target branch for fixing the failing tests

PR #802, opened 2024-05-02
replace `callsOnly` by`onlyCalls` in README

clemsos:
## What it solves

There is a mistake in the README of the Safe protocol-kit where the args for `createTransaction` specifies a `callsOnly` param - while it should be`onlyCalls`.

You can see the [types](https://github.com/safe-global/safe-core-sdk/blob/0bc0515d4265ef099085644791d4fa1fe4e1cf55/packages/protocol-kit/src/types/index.ts#L136) here 

## How this PR fixes it

This PR corrects the error in the README.

Also related : https://github.com/safe-global/safe-docs/pull/459


clemsos:

PR #804, opened 2024-05-03
[api kit] Update delegates endpoint to v2

### LINKED ISSUE #786: [api-kit] Update delegates endpoint
## Description

The API team is deprecating the v1 delegates endpoint and suggesting a new system to add, get and remove delegates.
The signature process is the biggest change.

We need to update to v2 delegates endpoint and support the new signing method they request. Currently is only available on staging.

## Proposed solution

 - Update getSafeDelegates
 - Update addSafeDelegate
 - Update removeSafeDelegate

Is possible that we don't need to change the API of the current methods in the `api-kit` as the signing is managed internally.

## Additional context

 - [API Kit PR](https://github.com/safe-global/safe-transaction-service/pull/2001)
 - [Staging swagger](https://safe-transaction-sepolia.staging.5afe.dev/)

leonardotc:
## What it solves
Resolves #786

## How this PR fixes it
- Updates the endpoints to use V2. 
- Change any necessary that for that. 
- Change the signing to work in tandem with eip712.


leonardotc:
I had to use this endpoint on the e2e tests: https://safe-transaction-sepolia.staging.5afe.dev/api

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
```diff
@@ -125,10 +126,11 @@ describe('addSafeDelegate', () => {
 
   it('should fail if the signer is not an owner of the Safe', async () => {
     const { safeApiKit, signer } = await getServiceClient(
-      '0xb0057716d5917badaf911b193b12b910811c1497b5bada8d7711f758981c3773'
+      '0xb0057716d5917badaf911b193b12b910811c1497b5bada8d7711f758981c3773',
+      'https://safe-transaction-sepolia.staging.5afe.dev/api'
```
yagopv, line 130: This should be removed before merging right? We will use the prod endpoint for the final version

packages/api-kit/tests/utils/signDelegate.ts
```diff
@@ -0,0 +1,22 @@
+import { Signer } from 'ethers'
+import { padHex } from '@safe-global/api-kit/utils/data'
+
+export async function signDelegate(signer: Signer, delegateAddress: string, chainId: string) {
```
yagopv, line 4: If this is only a test util perhaps make more sense to add it as a real util in the src/utils folder and use it both in the private method and in the tests. This allow only maintance in 1 place instead 2. WDYT?

leonardotc: Agreed, that was the initial approach. I forgot why I changed.

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
leonardotc: Great question. From what I understood, we will merge and then update.

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -316,9 +318,8 @@ class SafeApiKit {
     }
     const { address: delegate } = this.#getEip3770Address(delegateAddress)
     const { address: delegator } = this.#getEip3770Address(delegatorAddress)
-    const totp = Math.floor(Date.now() / 1000 / 3600)
-    const data = delegate + totp
-    const signature = await signer.signMessage(data)
+    const signature = await signDelegate(signer, delegate, this.#chainId)
+
     const body: any = {
```
yagopv, line 323: This is SafeDelegateResponse right?

yagopv:
```suggestion
    const body: SafeDelegateResponse = {
```

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
yagopv: @dasanra ?

packages/api-kit/src/SafeApiKit.ts
leonardotc: I think that is the request body which is marked as any on the type.

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
yagopv: We'll keep it for now

yagopv (approved):

PR #808, opened 2024-05-08
Passkey support

### LINKED ISSUE #789: [Passkeys] Add compatibility to the protocol-kit
## Context / issue

Passkey implementation is ready on protocol level. We need to add compatibility in the protocol-kit.
It should be possible to add an owner that is managed by a passkey.
This owner will be able to sign using EIP-1271.
The protocol-kit should be able to manage this signers

## Proposed solution

 - Make the protocol-kit to receive a passkey generated signer.
 - Add the passkey signer contracts to the protocol-kit.
 - Use the passkey signer to sign Safe transactions.
 - Add a passkey signer to a Safe.

## Additional context

 - [Research document.](https://www.notion.so/safe-global/Passkeys-Research-document-cacfbf24f65040a3a6926fc9d3ac5c40)

DaniSomoza:
## What it solves
Resolves #789 & #793 by adding support for passkeys in the `protocol-kit`.

### `protocol-kit` instantiation using passkeys

You can create an instance of the `protocol-kit` using a passkey as a Signer. The passkey Signer is a contract compliant with EIP-1271 standards.

the passkey argument is an object with 2 keys:
- `rawId`:  The raw identifier of the passkey. See [`rawId` docs](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/rawId) 
- `publicKey`: The passkey’s public key, generated via the getPublicKey() method, necessary for generating X & Y coordinates. See [`getPublicKey()` method docs](https://developer.mozilla.org/en-US/docs/Web/API/AuthenticatorAttestationResponse/getPublicKey)

```typescript

  const passkeyCredential = await navigator.credentials.create({ ... });

  const rawId = passkeyCredential.rawId;
  const publicKey = passkeyCredential.response.getPublicKey();

  const passkey = {
    rawId,
    publicKey
  };

  const safeWithPasskey = await Safe.create({
    provider: RPC_URL,
    signer: passkey,
    safeAddress
  });

```

### Add a passkey owner

Now you can use the `createAddOwnerTx` method present in the protocol-kit using a passkey:

```typescript
  const safeTransaction = await safe.createAddOwnerTx({
    passkey,
  });

```

This function returns the Safe transaction to add a passkey as owner and optionally change the threshold. If the passkey Signer contract has not been deployed, this function creates a batch that includes both the Signer contract deployment and the addOwner transaction.

### `isOwner` function

The isOwner function now checks if a specific address or passkey is an owner of the current Safe:

```typescript
  const safe = await Safe.create({
    provider: RPC_URL,
    safeAddress,
  });

  await safe.isOwner(passkey); // returns true or false

```

### Sign a transaction using passkeys

```typescript
  const safeWithPasskey = await Safe.create({
    provider: RPC_URL,
    signer: passkey,
    safeAddress,
  });

  const dumpTransaction = {
    to: safeAddress,
    value: "0",
    data: "0x",
  };

  const safeTransaction = await safeWithPasskey.createTransaction({
    transactions: [dumpTransaction],
  });

  const signedTransaction = await safeWithPasskey.signTransaction(
    safeTransaction
  );


// execution need to be handled via EOA

```

### Added the WebAuthn Signer Factory Contract

This `SafeWebAuthnSignerFactory` contract is used to create a Signer and to get the address of a Signer.


packages/protocol-kit/src/types/passkeys.ts
```diff
@@ -0,0 +1,9 @@
+export type PasskeyCoordinates = {
+  x: string
+  y: string
+}
+
+export type passkeyArgType = {
```
yagopv, line 6: Uppercase?

packages/protocol-kit/src/Safe.ts
```diff
@@ -114,10 +120,38 @@ class Safe {
...
+    const isPasskeySigner = signer && typeof signer !== 'string'
+
+    if (isPasskeySigner) {
+      const safeProvider = new SafeProvider({
+        provider
+      })
+      const chainId = await safeProvider.getChainId()
+      const customContracts = contractNetworks?.[chainId.toString()]
+
+      const safeWebAuthnSignerFactoryContract = await getSafeWebAuthnSignerFactoryContract({
```
yagopv, line 132: This can be retrieved from the safeProvider right?

packages/protocol-kit/src/types/safeConfig.ts
```diff
@@ -40,7 +41,7 @@ type SafeConfigWithPredictedSafeProps = {
 
 export type SafeConfigProps = {
   provider: SafeProviderConfig['provider']
-  signer?: SafeProviderConfig['signer']
+  signer?: string | passkeyArgType
```
yagopv, line 44: Why this change? SafeProviderConfig signer should be the correct one right?

packages/protocol-kit/src/utils/passkeys/PasskeySigner.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  /**
+   * Safe WebAuthn signer factory Contract.
+   */
+  safeWebAuthnSignerFactoryContract: SafeWebAuthnSignerFactoryContractImplementationType
+
+  constructor(
+    passkeyRawId: ArrayBuffer,
+    coordinates: PasskeyCoordinates,
+    safeWebAuthnSignerFactoryContract: SafeWebAuthnSignerFactoryContractImplementationType,
```
yagopv, line 36: What about using the safeProvider as a param and retrieve the contract and external provider from there?

packages/protocol-kit/src/contracts/SafeWebAuthnSignerFactory/v1.4.1/SafeWebAuthnSignerFactoryContract_v1_4_1.ts
```diff
@@ -0,0 +1,73 @@
...
+  SafeVersion,
+  SafeWebAuthnSignerFactoryContract_v1_4_1_Abi,
+  SafeWebAuthnSignerFactoryContract_v1_4_1_Contract,
+  SafeWebAuthnSignerFactoryContract_v1_4_1_Function,
+  SafeWebAuthnSignerFactory_1_4_1_ContractArtifacts
+} from '@safe-global/safe-core-sdk-types'
+import SafeProvider from '@safe-global/protocol-kit/SafeProvider'
+
+/**
+ * SafeWebAuthnSignerFactoryContract_v1_4_1  is the implementation specific to the Safe Proxy Factory contract version 1.4.1.
```
tmjssz, line 12: The `Safe Proxy Factory` is mentioned in multiple comments here, let's change it to the correct name to avoid confusion

packages/protocol-kit/src/contracts/safeDeploymentContracts.ts
```diff
@@ -175,3 +176,25 @@ export async function getSimulateTxAccessorContract({
...
+
+export async function getSafeWebAuthnSignerFactoryContract({
+  safeProvider,
+  safeVersion,
+  customContracts
+}: GetContractInstanceProps): Promise<SafeWebAuthnSignerFactoryContractImplementationType> {
+  const safeWebAuthnSignerFactoryContract = await safeProvider.getSafeWebAuthnSignerFactoryContract(
+    {
+      safeVersion,
+      customContractAddress: customContracts?.simulateTxAccessorAddress,
```
tmjssz, line 188: Should be like this, right?
```suggestion
      customContractAddress: customContracts?.safeWebAuthnSignerFactoryAddress,
      customContractAbi: customContracts?.safeWebAuthnSignerFactoryAbi
```

yagopv (commented):
I think we need an utility (buildPasskey) to build the object used with `navigator.credentials.create`

```
{
      pubKeyCredParams: [
        {
          // ECDSA w/ SHA-256: https://datatracker.ietf.org/doc/html/rfc8152#section-8.1
          alg: -7,
          type: 'public-key',
        },
      ],
      challenge: crypto.getRandomValues(new Uint8Array(32)),
      rp: {
        name: 'Safe Wallet',
      },
      user: {
        displayName: 'Safe Owner',
        id: crypto.getRandomValues(new Uint8Array(32)),
        name: 'safe-owner',
      },
      timeout: 60000,
      attestation: 'none',
    }
```

packages/protocol-kit/src/Safe.ts
```diff
@@ -421,13 +455,37 @@ class Safe {
...
   }
 
+  /**
+   * Checks if a specific address or passkey is an owner of the current Safe.
+   *
+   * @param owner - The owner address or a passkey object
+   * @returns TRUE if the account is an owner
+   */
+  async isOwner(owner: string | passkeyArgType): Promise<boolean> {
+    const isOwnerAddress = typeof owner === 'string'
```
leonardotc, line 465: The `this.#safeProvider` has an util to detect if this is an `address`. Consider to just query it. I dont think it will match eip3770 though.

packages/protocol-kit/src/Safe.ts
```diff
@@ -421,13 +455,37 @@ class Safe {
...
+      return this.#isOwnerAddress(owner)
+    }
+
+    // passkey flow
+    const webAuthnSignerFactoryContract = this.#contractManager.safeWebAuthnSignerFactoryContract
+    const provider = this.#safeProvider.getExternalProvider()
+
+    const passkeySigner = await PasskeySigner.init(owner, webAuthnSignerFactoryContract, provider)
+
+    const ownerAddress = await passkeySigner.getAddress()
```
leonardotc, line 477: I may be wrong but isnt all this chunk just the `this.#safeProvider.getSignerAddress()`?

packages/protocol-kit/src/Safe.ts
```diff
@@ -114,10 +120,38 @@ class Safe {
   async #initializeProtocolKit(config: SafeConfig) {
     const { provider, signer, isL1SafeSingleton, contractNetworks } = config
 
-    this.#safeProvider = new SafeProvider({
-      provider,
-      signer
-    })
+    const isPasskeySigner = signer && typeof signer !== 'string'
```
leonardotc, line 123: Disclaimer: The comment has nothing  to do with this line but wasnt the protocol kit supposed to have a `getPasskeyCreationOptions` or something like that for the people to have a schema on how to create a "safe-compliant" passkey or something?

packages/protocol-kit/src/SafeFactory.ts
```diff
@@ -62,7 +63,38 @@ class SafeFactory {
   }: SafeFactoryInitConfig) {
     this.#provider = provider
     this.#signer = signer
-    this.#safeProvider = new SafeProvider({ provider, signer })
+    const isPasskeySigner = signer && typeof signer !== 'string'
```
leonardotc, line 66: Consider on extracting this to a local util since its reused.

packages/protocol-kit/src/types/safeProvider.ts
```diff
@@ -11,12 +14,12 @@ export type HexAddress = string
...
 export type HttpTransport = string
 export type SocketTransport = string
-export type SafeSigner = HexAddress | PrivateKey
+export type SafeSigner = HexAddress | PrivateKey | PasskeySigner
 
 export type SafeProviderConfig = {
   /** signerOrProvider - Ethers signer or provider */
   provider: Eip1193Provider | HttpTransport | SocketTransport
-  signer?: HexAddress | PrivateKey
+  signer?: HexAddress | PrivateKey | passkeyArgType
```
leonardotc, line 22: Isnt this the `SafeSigner` above?

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -56,14 +65,20 @@ class SafeProvider {
...
-    if (this.signer && !ethers.isAddress(this.signer)) {
-      const privateKeySigner = new ethers.Wallet(this.signer, this.#externalProvider)
-      return privateKeySigner
-    }
-
-    if (this.signer) {
-      return this.#externalProvider.getSigner(this.signer)
+    if (typeof this.signer === 'string') {
+      // If the signer is not an Ethereum address, it should be a private key
+      if (this.signer && !ethers.isAddress(this.signer)) {
```
leonardotc, line 70: I think if it reached this point, we can assume that `this.signer` is defined. Please, remove the left-hand side conditionhere and below.

packages/protocol-kit/src/contracts/SafeWebAuthnSignerFactory/v1.4.1/SafeWebAuthnSignerFactoryContract_v1_4_1.ts
```diff
@@ -0,0 +1,73 @@
...
+    safeProvider: SafeProvider,
+    customContractAddress?: string,
+    customContractAbi?: SafeWebAuthnSignerFactoryContract_v1_4_1_Abi
+  ) {
+    const safeVersion = '1.4.1'
+    const defaultAbi = SafeWebAuthnSignerFactory_1_4_1_ContractArtifacts.abi
+
+    super(chainId, safeProvider, defaultAbi, safeVersion, customContractAddress, customContractAbi)
+
+    this.safeVersion = safeVersion
```
leonardotc, line 44: Any particular reason why its not just `this.safeVersion = '1.4.1'`?

packages/protocol-kit/src/utils/passkeys/PasskeySigner.ts
```diff
@@ -0,0 +1,238 @@
...
+  async sign(data: Uint8Array): Promise<string> {
+    const assertion = (await navigator.credentials.get({
+      publicKey: {
+        challenge: data,
+        allowCredentials: [{ type: 'public-key', id: this.passkeyRawId }],
+        userVerification: 'required'
+      }
+    })) as PublicKeyCredential & { response: AuthenticatorAssertionResponse }
+
+    if (!assertion || !assertion?.response?.authenticatorData) {
```
leonardotc, line 101: I dont think you need the first verification since you are using safe navigation/elvis operator.

packages/protocol-kit/src/utils/passkeys/PasskeySigner.ts
```diff
@@ -0,0 +1,238 @@
...
+  signMessage(message: string | Uint8Array): Promise<string> {
+    if (typeof message === 'string') {
+      return this.sign(ethers.getBytes(message))
+    }
+
+    return this.sign(message)
+  }
+
+  signTypedData(): Promise<string> {
+    throw new Error('Passkey Signers cannot sign signTypedData, they can only sign data.')
```
leonardotc, line 139: Consider on changing the `isTypedDataSigner` as well (although that will probably change the error).

packages/account-abstraction-kit/src/AccountAbstraction.ts
```diff
@@ -31,7 +31,10 @@ class AccountAbstraction {
   }
 
   #initializeProtocolKit = async () => {
-    const safeProvider = new SafeProvider({ provider: this.#provider, signer: this.#signer })
+    const safeProvider = new SafeProvider({
+      provider: this.#provider,
+      signer: this.#signer as string
```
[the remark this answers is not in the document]
DaniSomoza: done! I created a[ `createSafeProvider`](https://github.com/safe-global/safe-core-sdk/pull/841/files#diff-6800aef51d104476f1848ec54d4786a4ab2548533edd3ea5c0c584fbfbf032edR16) util function to create a proper `SafeProvider` object based on the `signer` provided by the user.

I added this in the other PR ([passkey-4337-support](https://github.com/safe-global/safe-core-sdk/pull/841)) to be able to use it also in the relay-kit. With this change the protocol-kit, the relay-kit and the account-abstraction-kit is compatible with passkeys.

packages/protocol-kit/src/Safe.ts
```diff
@@ -29,6 +29,7 @@ import ModuleManager from './managers/moduleManager'
 import OwnerManager from './managers/ownerManager'
 import {
   AddOwnerTxParams,
+  AddPasskeyOwnerTxParams,
```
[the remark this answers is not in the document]
DaniSomoza: I added two separate types to create guards:

```typescript
export function isAddOwnerTxParams(
  params: AddOwnerTxParams | AddPasskeyOwnerTxParams
): params is AddOwnerTxParams {
  return (params as AddOwnerTxParams).ownerAddress !== undefined
}

export function isAddPasskeyOwnerTxParams(
  params: AddOwnerTxParams | AddPasskeyOwnerTxParams
): params is AddPasskeyOwnerTxParams {
  return (params as AddPasskeyOwnerTxParams).passkey !== undefined
}

````

File: packages/protocol-kit/src/Safe.ts [hunk 18]
```diff
@@ -38,6 +39,7 @@ import {
   SigningMethod,
   SigningMethodType,
   SwapOwnerTxParams,
+  passkeyArgType,
```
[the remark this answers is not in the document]
DaniSomoza: done!

packages/protocol-kit/src/Safe.ts
DaniSomoza: yep, but I updated this code to use the `createSafeProvider` util function to wrap all the logic about to create the `safeProvider` object to reuse it in all packages.

packages/protocol-kit/src/types/safeConfig.ts
DaniSomoza: true! I restored it in the other PR:  https://github.com/safe-global/safe-core-sdk/pull/841/files#diff-d7c347ab5fed07174eb38d1a9731c90191e250ca1afabe36c35924940823a8cbR44

packages/protocol-kit/src/utils/passkeys/PasskeySigner.ts
DaniSomoza: I think that the best option is to avoid this, because the user can use the custom contracts and we have all this logic in the `Safe.init()` function.

So to avoid to include this `customContracts` logic inside of the SafeProvider, I think that we should keep this:

```typescript

    const safeWebAuthnSignerFactoryContract = await getSafeWebAuthnSignerFactoryContract({
      safeProvider,
      safeVersion: '1.4.1',
      customContracts
    })

    const passkeySigner = await PasskeySigner.init(
      signer,
      safeWebAuthnSignerFactoryContract,
      safeProvider.getExternalProvider()
    )
    
     new SafeProvider({
      provider,
      signer: passkeySigner
    })
```

packages/protocol-kit/src/contracts/SafeWebAuthnSignerFactory/v1.4.1/SafeWebAuthnSignerFactoryContract_v1_4_1.ts
DaniSomoza: done!

packages/protocol-kit/src/contracts/safeDeploymentContracts.ts
DaniSomoza: done!

packages/protocol-kit/src/Safe.ts
DaniSomoza: In this case we need to know if the `owner` param is a `string` or a `passkeyArgType` not an `address`.

packages/protocol-kit/src/Safe.ts
DaniSomoza: the `this.#safeProvider` might have been initialized with an EOA instead of a passkey. We need to ensure that the owner's address is created from the passkey provided by the user in the `isOwner` function parameter, not from the one in the `this.#safeProvider`.

packages/protocol-kit/src/SafeFactory.ts
DaniSomoza: done! I addressed this comment in [this other PR](https://github.com/safe-global/safe-core-sdk/pull/841/). I created a util in called [`createSafeProvider`](https://github.com/safe-global/safe-core-sdk/pull/841/files#diff-6800aef51d104476f1848ec54d4786a4ab2548533edd3ea5c0c584fbfbf032edR16) to reuse all the logic behind this:

Captura de pantalla 2024-06-04 a las 19 20 29

packages/protocol-kit/src/types/safeProvider.ts
DaniSomoza: nope, in the `SafeSigner` the `PasskeySigner` is a different type

packages/protocol-kit/src/SafeProvider.ts
DaniSomoza: done!

packages/protocol-kit/src/contracts/SafeWebAuthnSignerFactory/v1.4.1/SafeWebAuthnSignerFactoryContract_v1_4_1.ts
DaniSomoza: It is because `super(...)` call must be the first statement in the constructor. after `super(...)` call we can set the `this.safeVersion` value

packages/protocol-kit/src/utils/passkeys/PasskeySigner.ts
DaniSomoza: done!

packages/protocol-kit/src/types/safeFactory.ts
```diff
@@ -24,7 +25,7 @@ export type SafeFactoryConfig = {
 
 export type SafeFactoryInitConfig = {
   provider: SafeProviderConfig['provider']
-  signer?: SafeProviderConfig['signer']
+  signer?: string | PasskeyArgType
```
dasanra, line 28: This is a regression, right?

Should be as it was

```suggestion
  signer?: SafeProviderConfig['signer']
```

As the `SafeProviderConfig['signer']` already includes the `PasskeyArgType`

DaniSomoza: yes! fixed here:

```
https://github.com/safe-global/safe-core-sdk/pull/841/commits/801f8f237f9da2dc1a1adc7d4f844c79d46e5296
```

packages/protocol-kit/src/SafeFactory.ts
DaniSomoza: After your suggestion[ in this comment](https://github.com/safe-global/safe-core-sdk/pull/841#discussion_r1627310306), I create an `static async init` method in the `SafeProvider` instead of the [`createSafeProvider`](https://github.com/safe-global/safe-core-sdk/pull/841/files#diff-6800aef51d104476f1848ec54d4786a4ab2548533edd3ea5c0c584fbfbf032edR16)

packages/protocol-kit/src/types/safeProvider.ts
```diff
@@ -11,12 +14,12 @@ export type HexAddress = string
...
 export type HttpTransport = string
 export type SocketTransport = string
-export type SafeSigner = HexAddress | PrivateKey
+export type SafeSigner = HexAddress | PrivateKey | PasskeySigner
 
 export type SafeProviderConfig = {
   /** signerOrProvider - Ethers signer or provider */
   provider: Eip1193Provider | HttpTransport | SocketTransport
-  signer?: HexAddress | PrivateKey
+  signer?: HexAddress | PrivateKey | PasskeySigner | PasskeyArgType
```
yagopv, line 22: Confused here with the added `PasskeySigner`. This is to instantiate the provider with something from outside as an address, a privateKey or a passkey args object. Can this be instantiate with the PasskeySigner class?

In any case this should be added as weel to the SafeSigner type above that now is being used in multiple places. I would prefer to continue using SafeProviderConfig['signer'] as well

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -23,24 +30,34 @@ import {
...
+import { DEFAULT_SAFE_VERSION } from './contracts/config'
 
 class SafeProvider {
   #externalProvider: BrowserProvider | JsonRpcProvider
-  signer?: string
-  provider: Eip1193Provider | HttpTransport | SocketTransport
-
-  constructor({ provider, signer }: SafeProviderConfig) {
+  provider: SafeProviderConfig['provider']
+  signer?: SafeSigner
```
yagopv, line 52: I would use the SafeProviderConfig['signer'] in all places and remove the SafeSigner type or add a new SafeEthereumProvider (for example) instead so we would use:

```
provider: SafeProviderConfig['provider'],
signer: SafeProviderConfig['signer']
```

or

```
provider: SafeEthereumProvider,
signer: SafeSigner
```

yagopv: This is only to make it cleaner as this two props are used in multiple places

dasanra (approved):

PR #815, opened 2024-05-13
chore: add CONTRIBUTING and SUPPORT files

dasanra:
## What it solves
Adds CONTRIBUTING and SUPPORT files in order to guide developers how to use this repository.

Adds many more links to the docs so the reference is easier to find.

 - In the CONTRIBUTING we ask devs not to open issues for usability questions and redirect them to the SUPPORT file.
 - In the SUPPORT file we explain that the main source of information after the docs should be Stack Exchange and explain how to add a question there.
 - In the CONTRIBUTING we explain how to use the monorepo (install dependencies, build the project...)
 - Link CONTRIBUTING and SUPPORT on each package README.
 - Remove API reference from `protocol-kit` and `api-kit` README


CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+- [I Have a Question](#i-have-a-question)
+- [I Want to Contribute](#i-want-to-contribute)
+
+## <a name="i-have-a-question">I Have a Question</a>
+
+See our [support guidelines](https://github.com/safe-global/safe-core-sdk/tree/main/SUPPORT.md). **Do not** use GitHub issues for general support or questions.
+
+## <a name="i-want-to-contribute">I Want to Contribute</a>
+### Legal Notice
+You will need to agree to [our CLA](https://safe.global/cla) in order to be possible to consider your contribution.
```
yagopv, line 16: Should we explain how to agree ?

dasanra: I think is enough with the bot instructions. If it wasn't enough then it shold be better to update the action to print a longer message. Here it will just add complexity to the document, while I thought it was important for them to know they should agree to a CLA before doing the effort to create the code

guides/README.md
```diff
@@ -6,7 +6,7 @@ Read about the basics of Safe and how it compares to other solutions [here](http
 
 ## Safe Core SDK
```
yagopv, line 7: Not sure if this guide make sense instead using the docs.

In any case I will do one of the following with the `integrating-the-safe-core-sdk.md` guide and corresponding

1) Update the guide for reflecting our usual naming:
- safeService -> apiKit
- safedk -> protocolKit
- Safe.create -> Safe.init
2) Remove README and guide as the docs are probably covering the same info

packages/api-kit/README.md
```diff
@@ -8,46 +8,27 @@ Software development kit that facilitates the interaction with the [Safe Transac
...
-yarn test
-npm run test
+yarn add @safe-global/api-kit
+npm install @safe-global/api-kit
 ```
 
-## <a name="initialization">Initialization</a>
-
-### Initialize the SafeApiKit
+## Quick Start
```
yagopv, line 31: Below i will update the safeService const to apiKit

packages/protocol-kit/README.md
```diff
@@ -4,20 +4,22 @@
 [![GitHub Release](https://img.shields.io/github/release/safe-global/safe-core-sdk.svg?style=flat)](https://github.com/safe-global/safe-core-sdk/releases)
 [![GitHub](https://img.shields.io/github/license/safe-global/safe-core-sdk)](https://github.com/safe-global/safe-core-sdk/blob/main/LICENSE.md)
 
-Software development kit that facilitates the interaction with the [Safe contracts](https://github.com/safe-global/safe-contracts).
+Software development kit that facilitates the interaction with the [Safe Smart Account](https://github.com/safe-global/safe-smart-account).
```
yagopv, line 7:
```suggestion
Software development kit that facilitates the interaction with [Safe Smart Accounts](https://github.com/safe-global/safe-smart-account).
```

packages/api-kit/README.md
dasanra: Done

guides/README.md
dasanra: I went to option 1 for now. I would still leave the guide around as it's updated to the latest changes.

We don't have such guide in the docs yet so I would leave it here as an example that we can port later.

README.md
```diff
@@ -21,6 +38,18 @@ Software developer tools that facilitate the interaction with the Safe [contract
 | [relay-kit](https://github.com/safe-global/safe-core-sdk/tree/main/packages/relay-kit)                       | ​​​[​![NPM Version](https://badge.fury.io/js/%40safe-global%2Frelay-kit.svg)​](https://badge.fury.io/js/%40safe-global%2Frelay-kit)​             | Library to abstract transaction fees payment (gas fees), allowing the use of native tokens or ERC-20​​                                                                                 |
```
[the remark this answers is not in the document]
dasanra: Done

yagopv (approved):

PR #834, opened 2024-05-29
776 addv7compatibility

### LINKED ISSUE #845: [4337] Support 0.7 Entrypoint implementation
## Context / issue

0.7 entrypoint was unfinished when we started 4337 implementation. Currently it seems ready to be used and it should be possible to use it with the Safe SDK. We need to test and support both versions in the SDK.

## Proposed solution

 - Update the `safe-modules-deployments`to the latest version, having compatibility for 0.2.0 or 0.3.0 versions of the Safe module.
 - We should keep 0.6 entrypoint by default (which is supported by 0.2.0 version of the Safe module). The API is not supporting v0.7 UserOperations yet.
 - From the supported entrypoints we receive from the bundler we need to detect which one is 0.6 or 0.7.
 - When we load a Safe that is already deployed, the SDK needs to be able to select the correct module that the Safe is using.
 - `AddModulesLib` was renamed to `SafeModuleSetup`. Everything will still work if unchanged, but would be good to align naming convention to avoid future breaking changes.
 - Add tests that ensure we support both versions.

## Additional context

[safe-modules CHANGELOG](https://github.com/safe-global/safe-modules/blob/main/modules/4337/CHANGELOG.md#version-030)

The API service won't be compatible with 0.7.0 yet. We can add compatibility to the SDK setting the specific module address manually. We should test this approach to anticipate the differences in the SafeOperation object.

leonardotc:
## What it solves
Resolves #845


dasanra:
Many breaking changes affecting this PR. We will start from raw when task is prioritized.

Issue #825, opened 2024-05-27
[4337] Add compatibility to convert API received Operations


dasanra:
## Context / issue

In the protocol-kit, those methods which make sense that receive a transaction from the API can receive either the SDK object or the API object (signTransaction, executeTransaction...) and they detect the type of object converting internally to something they can use. We are not doing this kind of conversion in the 4337 kit, which makes it unconvenient to use.

## Proposed solution

Allow the sign and execute methods in Safe4337Pack to receive either the SDK object or the API object. If it's the API response, it should be able to convert to an object which is able to use.

We should also make sure that the SDK returned object is already compatible with the API consumed object.

PR #836, opened 2024-05-30
feat(relay-kit): Add tx service compatibility to `Safe4337Pack`

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/825

## How this PR fixes it
We are adding support for using SafeOperation responses from the transaction service in the `Safe4337Pack`

### DIFF
--- a/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
+++ b/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
@@ -405,6 +405,7 @@ export class Safe4337Pack extends RelayKitBasePack<{
     }
 
     const safeOperation = new EthSafeOperation(userOperation, {
+      moduleAddress: this.#SAFE_4337_MODULE_ADDRESS,
       entryPoint: this.#ENTRYPOINT_ADDRESS,
       validUntil,
       validAfter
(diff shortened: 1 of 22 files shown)


packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -476,9 +520,15 @@ export class Safe4337Pack extends RelayKitBasePack<{
    * @param {EthSafeOperation} safeOperation - The SafeOperation to execute.
    * @return {Promise<string>} The user operation hash.
    */
-  async executeTransaction({
-    executable: safeOperation
-  }: Safe4337ExecutableProps): Promise<string> {
+  async executeTransaction({ executable }: Safe4337ExecutableProps): Promise<string> {
+    let safeOperation: EthSafeOperation
+
+    if ('safeOperationHash' in executable) {
```
tmjssz, line 526: How about defining this as a type guard, so you could use it as `!isSafeOperation(executable)` here, as well as in line 467. You could even use it in [SafeApiKit line 795](https://github.com/safe-global/safe-core-sdk/blob/feat/add-api-compatibility-to-4337-pack/packages/api-kit/src/SafeApiKit.ts#L795)

yagopv: Hey, changed. Defined the guards in the types library together with the type definition

tmjssz (approved):

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -415,17 +419,67 @@ export class Safe4337Pack extends RelayKitBasePack<{
...
+
+    if (safeOperationResponse.confirmations) {
+      safeOperationResponse.confirmations.forEach((confirmation: SafeOperationConfirmation) => {
+        safeOperation.addSignature(new EthSafeSignature(confirmation.owner, confirmation.signature))
+      })
+    }
+
+    return safeOperation
+  }
+
```
dasanra, line 461: Don't you feel we are adding lot of utils like functionality into the main class? We even have some other utils at the bottom of this file that makes the file to be very long.

yagopv: yes, you are right

dasanra (approved):

PR #838, opened 2024-05-30
feat(protocol-kit): Tests for passkey

### LINKED ISSUE #791: [Passkeys] Test protocol-kit
## Context / issue

Why is this feature needed?

In case your feature request related to a problem, please add clear and concise description of what the issue is.

## Proposed solution

Please add a clear and concise description of what you want to happen.

## Alternatives

Please add a clear and concise description of any alternative solutions or features you have considered.

## Additional context

Add any other context about the feature request here.

tmjssz:
## What it solves
Resolves #791

## How this PR fixes it
Adding e2e tests for protocol-kit's passkey support.


packages/protocol-kit/contracts/Deps_V1_0_0.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity >=0.5.0 <0.9.0;
```
yagopv, line 2: Why is this being updated?

tmjssz: For visibility, here a summary of what Yago and I discussed outside of github:

- Reason for the pragma change is that the `SafeWebAuthnSignerFactory` contract needs to be compiled with at least version `0.8.24`, so it had to be added to the hardhat config
- Version `0.8.24` still satisfied the pragma `<0.9.0`, so the compiler picked it for other contracts as well then (since it's the highest version defined in the config)
  - This made the compilation of the `CompatibilityFallbackHandler` contract fail
  - To ensure, that it is compiled with the same version as before, the pragma was changed

tmjssz: After checking again, I noticed that the problem described above does only occur for `CompatibilityFallbackHandler v1.3.0`. So the only necessary pragma change is the one to `<=0.8.0` in [Deps_V1_3_0.sol](https://github.com/safe-global/safe-core-sdk/pull/838/files#diff-ee24319abcf333072392780b73d837bb426e396cfa9a5b53b2b71831af2884fbR2). Reverted the other pragma changes.

dasanra (approved):

PR #841, opened 2024-05-30
Passkey 4337 support

### LINKED ISSUE #790: [Passkeys] Deploy a Safe with passkey signer using the relay-kit
## Context / issue

It should be possible to deploy a Safe with the 4337 pack when using a passkey as a signer.

## Proposed solution

The protocol-kit should already be compatible with the passkey signer before implementing the relay-kit compatibility. 

It should be possible to deploy a 4337 compatible Safe when using a passkey signer.
It should be possible to use a 4337 compatible Safe when using a passkey signer.

DaniSomoza:
## What it solves
Resolves #790 

## How this PR fixes it

Adds compatibility to the `relay-kit` to get a passkey signer.


packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -191,41 +214,90 @@ export class Safe4337Pack extends RelayKitBasePack<{
...
+      // third transaction: passkey support via shared signer SafeWebAuthnSharedSigner
+      // see: https://github.com/safe-global/safe-modules/blob/main/modules/passkey/contracts/4337/experimental/README.md
+      const isPasskeySigner = signer && typeof signer !== 'string'
+
+      if (isPasskeySigner) {
+        const safeProvider = new SafeProvider({
+          provider
+        })
+
+        const safeWebAuthnSignerFactoryContract = await getSafeWebAuthnSignerFactoryContract({
```
leonardotc, line 237: Shouldnt we validate if this doesnt return? I understand that this function and not by the deployments and that, if it reaches this point, it means that we are probably in Sepolia (since it requires the deployment of 4337) but will that always be true?

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -441,17 +514,31 @@ export class Safe4337Pack extends RelayKitBasePack<{
...
+
+      // SafeWebAuthnSharedSigner signature
+      signature = new EthSafeSignature(
+        SAFE_WEBAUTHN_SHARED_SIGNER_ADDRESS,
+        passkeySignature.data,
+        true
+      )
+    } else {
+      if (
+        signingMethod === SigningMethod.ETH_SIGN_TYPED_DATA_V4 ||
```
leonardotc, line 532: Consider:
```
signingMethod in [SigningMethod.ETH_SIGN_TYPED_DATA_V4, SigningMethod.ETH_SIGN_TYPED_DATA_V3, SigningMethod.ETH_SIGN_TYPED_DATA]
```

DaniSomoza: done!

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
DaniSomoza: mmm good catch!

The `getSafeWebAuthnSignerFactoryContract` function throws an error if the contract is not deployed in the current chain.

```typescript
  if (!isContractDeployed) {
    throw new Error('safeWebAuthnSignerFactory contract is not deployed on the current network')
  }
```

However, after some research, I've realized that the current implementation only works when the `WebAuthnSignerFactory` contract is deployed.  (now its only deployed in `Sepolia`, and the address hardcoded in the code...).

To address this, I've removed the `getSafeWebAuthnSignerFactoryContract` call from the `ContractManager` (used in the `protocol-kit` instantiation). We now only invoke `getSafeWebAuthnSignerFactoryContract` when we need that specific contract (only in the passkey flows) rather than always calling it as before. 

See this commit for further details:

```
https://github.com/safe-global/safe-core-sdk/pull/841/commits/dd0dddc91d7382528a1acdf6369adbb9cac1ab69
```

packages/protocol-kit/src/types/safeConfig.ts
```diff
@@ -68,7 +67,7 @@ type ConnectSafeConfigWithPredictedSafeProps = {
 
 type ConnectSafeConfigProps = {
   provider?: SafeProviderConfig['provider']
-  signer?: string | PasskeyArgType
+  signer?: SafeProviderConfig['signer']
   /** isL1SafeSingleton - Forces to use the Safe L1 version of the contract instead of the L2 version */
```
dasanra, line 71:
```suggestion
  signer?: SafeProviderConfig['signer']
```

packages/protocol-kit/src/types/safeFactory.ts
```diff
@@ -14,7 +13,7 @@ export type DeploySafeProps = {
 
 export type SafeFactoryConfig = {
   provider: SafeProviderConfig['provider']
-  signer?: string | PasskeyArgType
+  signer?: SafeProviderConfig['signer']
   /** safeVersion - Versions of the Safe deployed by this Factory contract */
```
dasanra, line 17:
```suggestion
  signer?: SafeProviderConfig['signer']
```

PR #846, opened 2024-06-10
feat(relay-kit): Tests for using passkey with 4337

### LINKED ISSUE #792: [Passkeys] Test relay-kit
## Context / issue

Testing deployment when using the relay-kit. Continues https://github.com/safe-global/safe-core-sdk/issues/790

## Proposed solution

Test the possible cases when deploying a Safe using a passkey and 4337 module.

tmjssz:
## What it solves
Resolves #792

## How this PR fixes it
Adding tests for using passkeys with the Safe4337Pack.


.github/workflows/sdk-test.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       - name: Test
         env:
           PRIVATE_KEY: ${{ secrets.TESTING_PRIVATE_KEY }}
+          PASSKEY_PRIVATE_KEY: ${{ secrets.TESTING_PASSKEY_PRIVATE_KEY }}
```
tmjssz, line 31: This variable still needs to be added the secrets for the repo (it seems I don't have the permission to do that).

packages/protocol-kit/tests/e2e/utils/passkey.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Creates a mock passkey for testing purposes.
+ * @param name User name used for passkey mock
+ * @param webAuthnCredentials WebAuthnCredentials instance used for testing.
+ * @returns Passkey arguments obect with rawId and publicKey.
+ */
+export async function createMockPasskey(
+  name: string,
+  webAuthnCredentials: WebAuthnCredentials
+): Promise<PasskeyArgType> {
+  const passkeyCredential = await webAuthnCredentials.create({
```
yagopv, line 16: I think the await is not necessary here

yagopv (commented):
There are a duplicated passkey(s).ts file. Perhaps a renaming?

tmjssz:
> There are a duplicated passkey(s).ts file. Perhaps a renaming?

Ahh yes, I missed that. It actually didn't happen due to renaming, but because Leo and me did both move the passkey utils out to another file. I just didn't notice the difference in the `passkey(s)` file name. I removed `passkey.ts` now.

packages/protocol-kit/tests/e2e/utils/passkey.ts
tmjssz: Ah yes true! This file is removed, but I [updated the other util file `passkeys.ts` accordingly](https://github.com/safe-global/safe-core-sdk/pull/846/commits/5f5fee6ac7197d0a69c150c56a4c64ad9c07027e).

dasanra (approved):

DaniSomoza (approved):

PR #855, opened 2024-06-10
feat(protocol-kit): Restrict passkeys tests according to safe version

### LINKED ISSUE #852: [Passkeys] Testing Safe version compatibility
## Context / issue

Currently we are not checking the Safe version when trying to use passkeys. Passkeys are compatible with Safes v1.3.0 or up. We need to show an error if a lower version is intended to be used. Also currently we are not version locking specific passkeys tests,

## Proposed solution

Check the Safe version when using passkeys and throw in those versions lower than 1.3.0

leonardotc:
## What it solves
Resolves #852 by scoping the tests to the correct version.

## How this PR fixes it
Restrict some of the tests to the version that is applicable only. Adds a test that expects failure for the versions that are incompatbile (if it doesnt exists). The actual verification of the feature on the kit was previously added.


leonardotc:
The current branch should be merged on the passkeys branch. Since its aligned with the same branch, it should have some tests failing.

packages/protocol-kit/tests/e2e/passkey.test.ts
```diff
@@ -127,296 +127,347 @@ describe('Passkey', () => {
     }
   })
 
-  describe('isOwner', async () => {
-    it('should fail if the Safe is not deployed', async () => {
+  describeif(safeVersionDeployed < '1.3.0')('init', async () => {
```
dasanra, line 130:
```suggestion
  describeif(safeVersionDeployed < '1.3.0')('for Safe versions lower than 1.3.0', async () => {
```

packages/protocol-kit/tests/e2e/passkey.test.ts
```diff
@@ -127,296 +127,347 @@ describe('Passkey', () => {
...
+    it('should fail for Safe versions lower then 1.3.0', async () => {
```
dasanra, line 131:
```suggestion
    it('should fail when loading with a Passkey', async () => {
```

packages/protocol-kit/tests/e2e/passkey.test.ts
```diff
@@ -127,296 +127,347 @@ describe('Passkey', () => {
...
+          })
+        )
+        .to.be.rejectedWith(
+          'Current version of the Safe does not support the Passkey signer functionality'
+        )
     })
+  })
 
-    it('should return false if an account is not an owner of the connected Safe', async () => {
+  describe('isOwner', async () => {
```
dasanra, line 154:
```suggestion
  describeif(safeVersionDeployed >= '1.3.0')('for Safe versions higher than 1.3.0', async () => {
```

packages/protocol-kit/tests/e2e/passkey.test.ts
dasanra, line 131:
```suggestion
  describeif(safeVersionDeployed < '1.3.0')('for Safe versions lower than 1.3.0', async () => {
    it('should fail when `init` with a Passkey', async () => {
```

dasanra, line 131:
```suggestion
    it('should fail for Safe versions lower than 1.3.0', async () => {
```

packages/protocol-kit/tests/e2e/passkey.test.ts
```diff
@@ -127,296 +127,347 @@ describe('Passkey', () => {
...
+    itif(safeVersionDeployed < '1.3.0')(
+      'should fail for a passkey argument of Safe <v1.3.0',
```
dasanra, line 156:
```suggestion
      'should fail for Safe versions lower than 1.3.0',
```

dasanra: I was thinking a bit about this test. The error is thrown by the `init` method so the truth is that we do twice the same test, as we check this in the previous one.

leonardotc: Yes. The reason for that is that I came up with the test without looking too much at the implementation at first. I just thought:
- A protocol-kit instance cannot be valid if there is a passkey signer on a version lower than 1.3.0
- A valid protocol-kit cannot handle a `isOwner` query with passkey as an argument.

packages/protocol-kit/tests/e2e/passkey.test.ts
leonardotc: That would prevent the execution of the parameter test though.

packages/protocol-kit/tests/e2e/passkey.test.ts
dasanra: Ok, after your latest changes then updating this test description would be enough

packages/protocol-kit/tests/e2e/safeProvider.test.ts
```diff
@@ -35,6 +54,22 @@ describe('Safe contracts', () => {
     }
   })
 
+  describe('init', async () => {
+    itif(safeVersionDeployed < '1.3.0')(
+      'should return an L1 Safe contract from safe-deployments',
```
dasanra, line 59: This description doesn't seem correct

dasanra:
```suggestion
      'should fail for a passkey signer and Safe <v1.3.0',
```

dasanra (approved):

PR #857, opened 2024-06-11
feat(relay-kit): add dummy signature as a passkey signature

### LINKED ISSUE #849: [Passkeys] 4337 gas estimation not working
## Context / issue

Passkeys + 4337 gas estimation is not working as expected when using sponsored transactions. We need to provide a dummy passkey signature in order to get the estimation to trigger the signing logic

## Proposed solution

It should be possible to use 4337 sponsored transactions when a passkey is used as a signer. As the gas gets estimated by the paymaster and it's not allowing to suggest any changes after, it needs to simulate the passkey flow correctly.

DaniSomoza:
## What it solves
Resolves #849 

## How this PR fixes it

Passkeys + 4337 gas estimation is not working as expected when using sponsored transactions. We add a dummy passkey signature to get a valid `verificationGasLimit` estimation.

We assume that all owners are passkeys.

### DIFF
--- a/packages/relay-kit/src/packs/safe-4337/utils.ts
+++ b/packages/relay-kit/src/packs/safe-4337/utils.ts
@@ -13,6 +13,7 @@ import {
 } from '@safe-global/protocol-kit'
 import { ethers } from 'ethers'
 import { EIP712_SAFE_OPERATION_TYPE, INTERFACES } from './constants'
+import { SAFE_WEBAUTHN_SHARED_SIGNER_ADDRESS } from './Safe4337Pack'
 
 /**
  * Gets the EIP-4337 bundler provider.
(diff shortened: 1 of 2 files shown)


packages/relay-kit/src/packs/safe-4337/utils.ts
```diff
@@ -13,6 +13,7 @@ import {
 } from '@safe-global/protocol-kit'
 import { ethers } from 'ethers'
 import { EIP712_SAFE_OPERATION_TYPE, INTERFACES } from './constants'
+import { SAFE_WEBAUTHN_SHARED_SIGNER_ADDRESS } from './Safe4337Pack'
```
dasanra, line 16: This generates a "Circular dependency" between files, should be added to `constants`

packages/relay-kit/src/packs/safe-4337/utils.ts
```diff
@@ -121,21 +122,48 @@ export function userOperationToHexValues(userOperation: UserOperation) {
 }
 
 /**
- * This method creates a dummy signature for the SafeOperation based the owners.
+ * Passkey Dummy client data JSON fields. This can be used for gas estimations, as it pads the fields enough
+ * to account for variations in WebAuthn implementations.
+ */
+export const DUMMY_CLIENT_DATA_FIELDS = [
+  `"origin":"http://safe.global"`,
```
dasanra, line 129: Would it be better to use https to be more realistic?

packages/relay-kit/src/packs/safe-4337/utils.ts
DaniSomoza: done!

PR #860, opened 2024-06-13
feat: add short name for zircuit-testnet

leonimella:
## What it solves
Adds short name for Zircuit Testnet:
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-48899.json

https://github.com/safe-global/safe-deployments/pull/640


dasanra:
Will be solved in https://github.com/safe-global/safe-core-sdk/pull/874

PR #873, opened 2024-06-18
feat: add short name for open-campus-codex

leonimella:
## What it solves
Adds short name for Open Campus Codex:
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-656476.json

https://github.com/safe-global/safe-deployments/pull/650


dasanra:
fixed by #910

PR #876, opened 2024-06-20
feat(api-kit): Support for SafeOperation confirmations endpoints

### LINKED ISSUE #856: [4337] Add signature method for api-kit
## Context / issue

The add signature endpoint was missing at the time of implementation. This is a very useful method that lets gather signatures on Safes compatible with 4337 which have multiple signers.

## Proposed solution

 [ ] Add the `confirmatons` for `safe-operations` endpoint

## Additional context

https://github.com/safe-global/safe-transaction-service/issues/2060

tmjssz:
## What it solves
Resolves #856

## How this PR fixes it
Extend the `api-kit` by two functions using the following endpoints of the transaction-service.

### getSafeOperationConfirmations
Returns the list of confirmations for a given a SafeOperation.

```
GET /v1/safe-operations/{safe_operation_hash}/confirmations/
```

### confirmSafeOperation
Adds a confirmation for a SafeOperation.

```
POST /v1/safe-operations/{safe_operation_hash}/confirmations/
```


packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -151,7 +156,8 @@ describe('Endpoint tests', () => {
         .to.be.eventually.deep.equals({ data: { success: true } })
       chai.expect(fetchData).to.have.been.calledWith({
         url: `${txServiceBaseUrl}/v1/multisig-transactions/${safeTxHash}/confirmations/`,
-        method: 'get'
+        method: 'post',
+        body: { signature: '0x' }
```
dasanra, line 160: Wasn't this failing before? 

Maybe it's better to pass the signature parameter defined above in the body?

packages/api-kit/tests/e2e/getSafeOperationsByAddress.test.ts
```diff
@@ -1,4 +1,4 @@
-import SafeApiKit from '@safe-global/api-kit'
+import SafeApiKit from '@safe-global/api-kit/index'
```
dasanra, line 1: Is it mandatory to add the `index`?

packages/api-kit/tests/endpoint/index.test.ts
tmjssz: Yes true, updated.
It was not failing before, because the assertion got fulfilled by the previous test case `getTransactionConfirmations`. However, after I added the `sinon.resetHistory()` in the afterEach (line 56), this started to fail as expected.

packages/api-kit/tests/e2e/getSafeOperationsByAddress.test.ts
tmjssz: Without it, I see this weird error:
Screenshot 2024-06-20 at 17 12 36

All the other files in the same folder are defining the same import path for the SafeApiKit. That's why I changed it accordingly here.

tmjssz:
Some tests related to SafeDelegates started to fail, but it seems unrelated to my changes because I'm getting the same errors on the development branch

Screenshot 2024-06-20 at 17 26 21

dasanra:
Don't worry about the tests, they are not related with this PR. Already checked with the API team. 

Related to this https://github.com/safe-global/safe-core-sdk/pull/879

packages/api-kit/src/types/safeTransactionServiceTypes.ts
```diff
@@ -325,3 +286,12 @@ export type AddSafeOperationProps = {
     validAfter?: number
   }
 }
+
+export type GetSafeOperationConfirmationListProps = {
```
leonardotc, line 290: It is a little bit weird to use `limit` and `offset` as a string since they are associated with record count in a database. I imagine that the interface will be something like `apiKit.getSafeOperationConfirmations({ safeOperationHash, limit: '10' })`.

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -852,6 +854,66 @@ class SafeApiKit {
...
+
+  /**
+   * Returns the list of confirmations for a given a SafeOperation.
+   *
+   * @param getSafeOperationConfirmationsProps - The parameters for fetching the list of confirmations
+   * @returns The list of confirmations
+   * @throws "Invalid SafeOperation hash"
+   * @throws "Invalid data"
+   */
+  async getSafeOperationConfirmations({
```
leonardotc, line 866: Please double check how we do this. This kind of looks like a getSafeOperationConfirmationsByHash with a different name as the hash itself is a part of a "perma-link". In this case, it would be common to separate the interface as `getSafeOperationConfirmations(hash, { limit, offset })` since that would be somehow associated with a domain model.

packages/api-kit/src/types/safeTransactionServiceTypes.ts
tmjssz: Yes good point. Changed the type to number

packages/api-kit/src/SafeApiKit.ts
tmjssz: Updated the function signature accordingly 👍

packages/api-kit/src/SafeApiKit.ts
```diff
@@ -282,11 +284,11 @@ class SafeApiKit {
     if (label) {
       url.searchParams.set('label', label)
     }
-    if (limit) {
-      url.searchParams.set('limit', limit)
+    if (limit != null) {
```
yagopv, line 287: Why this? what happen if it's undefined or false?

yagopv (approved):

packages/api-kit/src/SafeApiKit.ts
tmjssz: This is to allow the value being `0`. With the condition we had before, it would not set the search param. 
Now it would not be set for undefined either. For false it would be set. But the prop's type is number, so this should not happen.

PR #877, opened 2024-06-20
feat: Create new package `safe-kit`

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/867, https://github.com/safe-global/safe-core-sdk/issues/866

## How this PR fixes it
We are adding a new package, `safe-kit`, which exports a simple client, `createSafeClient`, that exposes a `send()` method to execute transactions through Safe, making the deployment necessary if required


packages/safe-kit/src/lib/sendAndDeployTransaction.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Sends a transaction using the SafeAccountClient client.
+ *
+ * @param {TransactionBase[]} transactions - An array of transactions to be sent.
+ * @param {TransactionOptions} options - Options for executing the transaction.
+ * @param {SafeAccountClient} safeClient - The SafeAccountClient client used to send the transaction.
+ * @returns {Promise<SafeClientTransactionResult>} A promise that resolves to the result of the transaction.
+ */
+export const sendAndDeployTransaction = async (
```
tmjssz, line 14: This function looks very similar to `sendTransaction`. Have you considered merging them into one function with something like a `deploySafe` parameter flag?

packages/safe-kit/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+export type SafeClientTransactionResult = {
+  chain: {
+    hash?: string | undefined
+  }
+  safeServices?: {
+    safeTxHash: string | undefined
+  }
+}
+
+export type SafeClient = {
```
tmjssz, line 36: Do we need this type? Can we not use the `SafeAccountCient` class as type directly? Or what was the idea behind defining the type separately and making the class implement it?

packages/safe-kit/src/SafeAccountClient.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    ;(protocolKit.isSafeDeployed as jest.Mock).mockResolvedValue(false)
+    ;(protocolKit.getThreshold as jest.Mock).mockResolvedValue(1)
+    ;(sendAndDeployTransaction as jest.Mock).mockResolvedValue(TRANSACTION_RESPONSE)
+
+    const result: SafeClientTransactionResult = await safeAccountClient.send(transactions, options)
+
+    expect(protocolKit.isSafeDeployed).toHaveBeenCalled()
+    expect(protocolKit.getThreshold).toHaveBeenCalled()
+    expect(sendAndDeployTransaction).toHaveBeenCalledWith(transactions, options, safeAccountClient)
+    expect(result).toEqual({ chain: { hash: '0xTxHash' } })
```
tmjssz, line 60:
```suggestion
    expect(result).toEqual(TRANSACTION_RESPONSE)
```

packages/safe-kit/src/SafeAccountClient.test.ts
```diff
@@ -0,0 +1,89 @@
...
+  })
+
+  it('should throw an error if Safe is not deployed and threshold is greater than 1', async () => {
+    const transactions: TransactionBase[] = []
+    const options: TransactionOptions = {}
+    ;(protocolKit.isSafeDeployed as jest.Mock).mockResolvedValue(false)
+    ;(protocolKit.getThreshold as jest.Mock).mockResolvedValue(2)
+
+    await expect(safeAccountClient.send(transactions, options)).rejects.toThrowError(
```
tmjssz, line 69: I'm seeing a deprecation notice for `toThrowError`. Might be better to use `toThrow` instead.

yagopv: yes! done

packages/safe-kit/src/types.ts
yagopv: hehe, no, typical thing you start implementing one way and end with other. Good suggestion

yagopv: changed

packages/safe-kit/src/lib/sendAndDeployTransaction.ts
yagopv: It's not that similar, only the creation and signing part. However, if I create a deploy() function, I should either send the transaction signed or perform the signing internally, so it will be not super clear the implementation inside. Apart from that i would like to keep all these function under the same interface `(tx, options, safeClient) => txResponse`. I will keep it like this for now and let's see how we can evolve them

tmjssz (approved):

PR #884, opened 2024-06-21
chore: remove deprecated ethereum utils lib

dasanra:
## What it solves
refactor to remove deprecated ethereum utils lib and replace with Viem functions.

## How this PR fixes it
- Get from Viem the utils to predict the Safe address.
- Get from Viem the utils to check if signature has a prefix.


packages/protocol-kit/src/contracts/utils.ts
```diff
@@ -339,24 +345,24 @@ export async function predictSafeAddress({
...
 
-  const initializer = await encodeSetupCallData({
+  const initializer = (await encodeSetupCallData({
     safeProvider,
     safeAccountConfig,
     safeContract,
     customContracts,
     customSafeVersion: safeVersion // it is more efficient if we provide the safeVersion manually
-  })
+  })) as `0x${string}`
```
leonardotc, line 354: as Hex

Issue #870, opened 2024-06-17
[SDK Refactoring] Update SDK code to use `viem`


yagopv:
## Context / issue
After the previous refactor to remove the adapter pattern from the SDK, the next step is to migrate the usage of the `ethers` library to `viem`. 

Viem is designed with performance in mind, providing  optimized code for interacting with the blockchain. This can result in faster execution times for certain blockchain operations.  It adopts as well a modular design, allowing developers to import only what they need. This can lead to smaller bundle sizes in web applications when compared to ﻿`ethers`, which tends to be more monolithic

## Proposed solution
Remove ethers from the whole repo and substitute it for viem. This includes:
- Source Code (Packages)

## Additional context
[Research document](https://www.notion.so/safe-global/SDK-Refactor-4e45fd5f454d4715a60e72d4854b9aa9)
[Related update tests issue](https://github.com/safe-global/safe-core-sdk/issues/871)

PR #887, opened 2024-06-27
feat: Viem migration

### LINKED ISSUE #856: [4337] Add signature method for api-kit

tmjssz:
## What it solves
Resolves issues #870 and #871.

## How this PR fixes it
This pull request migrates the entire repository to `viem`, removing `ethers` as a dependency. `ethers` is retained only as a devDependency for the test-utils in `api-kit` and `protocol-kit`.

### protocol-kit

* **SafeProvider**
  * `getExternalProvider` now returns a viem `PublicClient` (previously an ethers `Provider`).
  * `getExternalSigner` now returns a viem `WalletClient` (previously an ethers `AbstractSigner`).
  * `getTransaction` now returns a viem `Transaction` (previously an ethers `TransactionResponse`).
  * `encodeParameters` and `decodeParameters` now expect a `string` as the first parameter, containing the Human-readable ABI parameters (previously a `string[]`).
* All ABI parameters are now a viem `Abi` (previously `JsonFragment | JsonFragment[]`). This affects:
  * `defaultAbi` and `customContractAbi` in all Contract classes.
  * `customContractAbi` parameter in all `get...ContractInstance` functions and in the `GetContractProps` type.
* `getSafeProxyFactoryContractInstance` function's `signerOrProvider` parameter is now a viem `PublicClient`.

### api-kit

* `SafeApiKit`'s `addSafeDelegate` and `removeSafeDelegate` functions now expect a viem `WalletClient` as the signer prop (previously an ethers `Signer`).

### relay-kit

* `Safe4337Options` now expects a `bundlerClient` prop of viem's type `PublicClient` (previously ethers `JsonRpcProvider`).

### auth-kit

* `AuthKitEthereumProvider` type is now a viem `EIP1193Provider` (previously ethers `Eip1193Provider`). This new type restricts the RPC methods for EIP-1193, as opposed to the ethers type, which accepted any string as a method name.


packages/api-kit/tests/e2e/addSafeOperation.test.ts
```diff
@@ -1,15 +1,19 @@
 import chai from 'chai'
 import chaiAsPromised from 'chai-as-promised'
-import { ethers } from 'ethers'
 import sinon from 'sinon'
 import sinonChai from 'sinon-chai'
 import Safe from '@safe-global/protocol-kit'
 import SafeApiKit from '@safe-global/api-kit/index'
+import { getAddSafeOperationProps } from '@safe-global/api-kit/utils/safeOperation'
 import { Safe4337Pack } from '@safe-global/relay-kit'
+import * as viem from '@safe-global/relay-kit/node_modules/viem'
```
dasanra, line 9: This import may be wrong

packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -363,7 +369,7 @@ describe('Endpoint tests', () => {
         nonce: 1
       }
       const origin = 'Safe Core SDK: Safe API Kit'
-      const signerAddress = await signer.getAddress()
+      const signerAddress = await walletClient.account.address
```
dasanra, line 372: Unnecessary await?

packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -411,7 +417,7 @@ describe('Endpoint tests', () => {
...
```
dasanra, line 420: Unnecessary await?

packages/api-kit/tests/endpoint/index.test.ts
```diff
@@ -33,8 +35,12 @@ const eip3770TokenAddress = `${config.EIP_3770_PREFIX}:${tokenAddress}`
 const safeTxHash = '0x317834aea988fd3cfa54fd8b2be2c96b4fd70a14d8c9470a7110576b01e6480a'
 const safeOpHash = '0x8b1840745ec0a6288e868c6e285dadcfebd49e846d307610a9ccd97f445ace93'
 const txServiceBaseUrl = 'https://safe-transaction-sepolia.safe.global/api'
-const defaultProvider = getDefaultProvider(config.JSON_RPC)
-const signer = new Wallet(PRIVATE_KEY_1, defaultProvider)
+
+const walletClient = createWalletClient({
```
dasanra, line 39: If we call this signer, the file would need less changes.

packages/auth-kit/jest.setup.ts
```diff
@@ -0,0 +1,3 @@
+import { TextEncoder, TextDecoder } from 'util'
+
+Object.assign(global, { TextDecoder, TextEncoder })
```
dasanra, line 3: A comment on why this is needed would be helpful

packages/onramp-kit/src/packs/monerium/SafeMoneriumClient.ts
```diff
@@ -195,31 +190,23 @@ export class SafeMoneriumClient extends MoneriumClient {
...
-        }),
-        this.#safeProvider.call({
-          from: safeAddress,
-          to: safeAddress,
-          data: eip1271BytesData
+      const [eip1271data, eip1271BytesData] = [EIP_1271_ABI, EIP_1271_BYTES_ABI].map((abi) =>
+        encodeFunctionData({
+          abi,
+          functionName: 'isValidSignature',
+          args: [messageHash as Hash, '0x']
```
dasanra, line 197: One of the `encodeFunctionData` received the `messageHash`, and the other one received the `msgBytes`.

This simplification can't be done.

packages/onramp-kit/jest.setup.ts
```diff
@@ -0,0 +1,3 @@
+import { TextEncoder, TextDecoder } from 'util'
+
+Object.assign(global, { TextDecoder, TextEncoder })
```
dasanra, line 3: A comment on why this is needed would be helpful

packages/onramp-kit/package.json
```diff
@@ -41,7 +41,8 @@
     "@safe-global/safe-core-sdk-types": "^5.0.2",
     "@stripe/crypto": "^0.0.4",
     "@stripe/stripe-js": "^1.54.2",
-    "ethers": "^6.13.1"
+    "ethers": "^6.13.1",
```
dasanra, line 44: This package is still using `ethers`

packages/api-kit/tests/e2e/addSafeOperation.test.ts
tmjssz: I know the import looked weird but it was actually intentionally, because it fixed an error that I was getting with a regular `import viem from 'viem'` statement:
```
Error: Trying to stub property 'createPublicClient' of undefined
```

➡️ https://github.com/safe-global/safe-core-sdk/actions/runs/9856259113/job/27212869062?pr=887

packages/auth-kit/jest.setup.ts
tmjssz: Got this error before:
```
 FAIL  src/packs/monerium/MoneriumPack.test.ts
  ● Test suite failed to run

    ReferenceError: TextEncoder is not defined

    > 1 | import { Hash, encodeFunctionData, hashMessage } from 'viem'
```

TextEncoder is used by viem but not bundled with it. Therefore we need to polyfill it for the tests.

Also added this as a comment to the code.

packages/onramp-kit/src/packs/monerium/SafeMoneriumClient.ts
leonardotc: They should be the same. The previous code had a `getBytes` which should the same hash but in bytes. In abitype both `adressType` and `bytesType` have the same types which is a `0x`-prefixed `string`: https://github.com/wevm/abitype/blob/main/packages/abitype/src/register.ts#L132 which matches the viem type for Hash https://github.com/wevm/viem/blob/f43388495e6f87147108ea101b6083213c5699c0/src/types/misc.ts#L5. I do think its a bit misleading and `Hex` is probably a bit more accurate. I will also test out because surprises happen.

packages/api-kit/src/utils/signDelegate.ts
```diff
@@ -1,10 +1,14 @@
-import { Signer } from 'ethers'
+import { Chain, LocalAccount, Transport, WalletClient } from 'viem'
```
yagopv, line 1: Should we use Account instead LocalAccount in all the places?

An Account can be both LocalAccount | JsonRPCAccount

packages/api-kit/src/utils/signDelegate.ts
```diff
@@ -1,10 +1,14 @@
...
+export async function signDelegate(
+  walletClient: WalletClient<Transport, Chain, LocalAccount>,
+  delegateAddress: string,
+  chainId: bigint
+) {
   const domain = {
     name: 'Safe Transaction Service',
     version: '1.0',
-    chainId: chainId
+    chainId: Number(chainId)
```
yagopv, line 11: Was this failing before?

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
```diff
@@ -11,17 +13,21 @@ const PRIVATE_KEY_1 = '0x83a415ca62e11f5fa5567e98450d0f82ae19ff36ef876c10a8d448c
...
+      chain: sepolia,
+      transport: http(),
+      account: privateKeyToAccount(PRIVATE_KEY_1)
+    })
   })
 
   it('should fail if Label is empty', async () => {
     const delegateAddress = '0x9cCBDE03eDd71074ea9c49e413FA9CDfF16D263B'
-    const delegatorAddress = await signer.getAddress()
+    const delegatorAddress = signer.account!.address
```
yagopv, line 30: This can be undefined right? Why is better than something like:

```
const accounts = await walletClient.getAddresses() 
```

packages/api-kit/tests/e2e/addSafeOperation.test.ts
```diff
@@ -1,15 +1,19 @@
...
+import { getAddSafeOperationProps } from '@safe-global/api-kit/utils/safeOperation'
+import { BundlerClient, Safe4337Pack } from '@safe-global/relay-kit'
 import { generateTransferCallData } from '@safe-global/relay-kit/packs/safe-4337/testing-utils/helpers'
-import { RPC_4337_CALLS } from '@safe-global/relay-kit/packs/safe-4337/constants'
+import {
+  ENTRYPOINT_ABI,
+  ENTRYPOINT_ADDRESS_V06,
+  RPC_4337_CALLS
+} from '@safe-global/relay-kit/packs/safe-4337/constants'
+import * as safe4337Utils from '@safe-global/relay-kit/dist/src/packs/safe-4337/utils'
```
yagopv, line 15: We should not import from dist folder

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -32,49 +22,114 @@ import {
...
+  Chain
+} from 'viem'
+import { privateKeyToAccount } from 'viem/accounts'
+import {
+  toEstimateGasParameters,
+  toCallGasParameters,
+  sameString
+} from '@safe-global/protocol-kit/utils'
+
+function asBlockId(blockId: number | string | undefined) {
```
yagopv, line 52: Shouldn't be this utils in an utils file?

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -32,49 +22,114 @@ import {
...
+}
+
+function blockTag(blockTag: any) {
+  return { blockTag: blockTag as BlockTag }
+}
 
 class SafeProvider {
-  #externalProvider: BrowserProvider | JsonRpcProvider
+  #chain?: Chain
+  #externalProvider: PublicClient
```
yagopv, line 66: It could be a WalletClient as well right? If you want to sign or send something using the client i think with the PublicClient you cant

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -32,49 +22,114 @@ import {
...
 
   constructor({ provider, signer }: SafeProviderConfig) {
     if (typeof provider === 'string') {
-      this.#externalProvider = new JsonRpcProvider(provider)
+      this.#externalProvider = createPublicClient({
+        transport: http(provider)
+      })
     } else {
-      this.#externalProvider = new BrowserProvider(provider)
+      const client = createPublicClient({
```
yagopv, line 76: client variable is not really necessary here

packages/protocol-kit/src/SafeProvider.ts
```diff
@@ -32,49 +22,114 @@ import {
...
-  #externalProvider: BrowserProvider | JsonRpcProvider
+  #chain?: Chain
+  #externalProvider: PublicClient
   signer?: string
   provider: Eip1193Provider | HttpTransport | SocketTransport
 
   constructor({ provider, signer }: SafeProviderConfig) {
     if (typeof provider === 'string') {
-      this.#externalProvider = new JsonRpcProvider(provider)
+      this.#externalProvider = createPublicClient({
```
yagopv, line 72: I think perhaps this logic could be directly used in the getExternalProvider. But more than that, I think we can rename getExternalProvider and getExternalSigner to getPublicClient and getWalletClient as it is more viem friendly for the developers to understand what it returns WDYT?

packages/protocol-kit/src/contracts/BaseContract.ts
```diff
@@ -80,24 +98,68 @@ class BaseContract<ContractAbiType extends InterfaceAbi & Abi> {
...
+    return { chain, ...txOptions, account } // Needs to be in this order to override the `account` if necessary
+  }
+
+  getChain(): Chain | undefined {
+    return getChainById(this.chainId)
   }
 
   getAddress: GetAddressFunction = () => {
-    return this.contract.getAddress()
+    return Promise.resolve(this.contract.address)
```
yagopv, line 134: This is like forcing to be a Promise. If we have an address perhaps is not a function retuning a Promise anymore

packages/protocol-kit/src/contracts/BaseContract.ts
```diff
@@ -80,24 +98,68 @@ class BaseContract<ContractAbiType extends InterfaceAbi & Abi> {
...
+      client: this.wallet || this.runner
+    })
+  }
+
+  async getTransactionReceipt(hash: Hash) {
+    const client = this.runner
+    return client?.getTransactionReceipt({ hash })
+  }
+
+  async convertOptions(
```
yagopv, line 114: Why is this conversion necessary. Would be helpful to have some comments / jsdoc here or in the implementations inside the protocol-kit utils

packages/protocol-kit/src/contracts/SafeProxyFactory/v1.0.0/SafeProxyFactoryContract_v1_0_0.ts
```diff
@@ -38,7 +40,7 @@ class SafeProxyFactoryContract_v1_0_0
     safeProvider: SafeProvider,
     customContractAddress?: string,
     customContractAbi?: SafeProxyFactoryContract_v1_0_0_Abi,
-    runner?: ContractRunner | null
+    runner?: PublicClient
```
yagopv, line 43: General thing. We call this runner everywhere but it's something inherited from ethers isn't it?  Perhaps we can simply call it client

packages/safe-kit/src/utils/sendTransaction.ts
```diff
@@ -17,15 +17,53 @@ export const sendTransaction = async ({
...
+    account: signer.account
   })
 
-  const txResponse = await txResponsePromise.wait()
+  const receipt = await client.waitForTransactionReceipt({ hash })
+
+  return receipt.transactionHash
+}
+
+export function createLegacyTxOptions(options?: TransactionOptions) {
```
yagopv, line 42: Have we got an util in the protocol-kit for this?

yagopv:
General question, did you retest the playgrounds?

packages/api-kit/src/utils/signDelegate.ts
tmjssz: No with ethers it was not failing, but viem's `signTypedData` function expects this value as number, instead of bigint.

packages/api-kit/src/utils/signDelegate.ts
tmjssz: Yes definitely!

packages/api-kit/tests/e2e/addSafeDelegate.test.ts
tmjssz: Yes that might be better. Updated

packages/api-kit/tests/e2e/addSafeOperation.test.ts
tmjssz: I had to import it like this in order to mock the `getEip4337BundlerProvider` function (see line 44). As this is a different package, the mock does only work when the dist version is used here.

packages/protocol-kit/src/SafeProvider.ts
tmjssz: Yes, done!

packages/protocol-kit/src/SafeProvider.ts
tmjssz: Updated

packages/protocol-kit/src/SafeProvider.ts
tmjssz: I think it would make sense, but it would also introduce additional breaking changes which we tried to keep at a minimum on this PR. So I think it would be better to do this in separate PR.

packages/protocol-kit/src/contracts/SafeProxyFactory/v1.0.0/SafeProxyFactoryContract_v1_0_0.ts
tmjssz: I like the idea, but it would probably introduce breaking changes as well. So I would prefer doing this in a follow-up task

packages/safe-kit/src/utils/sendTransaction.ts
tmjssz: Yes we do, but there was a reason why Leo didn't use it, he just didn't remember. We can check it again

tmjssz: The difference seems to be only that this function here additionally defines `maxFeePerGas` and `maxPriorityFeePerGas` compared to the one in `packages/protocol-kit/src/utils/transactions/utils.ts`. Maybe this was the reason

Screenshot 2024-07-25 at 14 48 29

tmjssz: After checking further, I could not find any use of the `createLegacyTxOptions` function defined here. Is it really needed or maybe just a forgotten leftover after a refactor? @leonardotc

packages/protocol-kit/src/contracts/BaseContract.ts
leonardotc: The conversion is necessary because the `writeContract` parameters takes a `TransactionRequest` type (with some omitted properties that can be inferred from the `Contract` instance. Viem does discriminate between transaction types. There is a specific type introduced post-london hardfork that changed some of the fee fields. What they do is that they create a union between a `BaseTransactionRequest` and the different fee properties as defined here: `https://github.com/wevm/viem/blob/viem%402.18.0/src/types/fee.ts`. Our `TransactionOptions` is somehow a more permissable one while their types are mutually exclusive. Our `TransactionOptions` are used in multiple places so I didnt want to break the consistency between the interface (and that its even destructured into the `TransactionResult` which is also all over the place. I added a comment to address that on the function as suggested.

packages/protocol-kit/src/SafeProvider.ts
tmjssz: Because this was a bigger change, I addressed it in a new PR: #928

packages/safe-kit/src/utils/sendTransaction.ts
leonardotc: Yeah, probably. I removed a lot of duplicates. I possibly forgot that one.

packages/protocol-kit/src/utils/eip-712/encode.ts
```diff
@@ -0,0 +1,211 @@
...
+  getTypesForEIP712Domain,
+  validateTypedData,
+  hashDomain,
+  toHex,
+  Hex,
+  HashTypedDataParameters
+} from 'viem'
+import { asHex } from '../types'
+
+// This whole file was copied (and softly adapted) from viem in order to expose the function that provides just the encoding.
```
dasanra, line 16: I saw this was done to avoid to break the types in the current iteration. This was not explicitly explained here, and that this piece of code should be removed as soon as possible. Please add some comment signaling this.

PR #888, opened 2024-06-28
Add hashSafeOperation to Safe4337Pack

bh2smith:
Currently the only way to get a SafeOpHash is to provide other details that are already part of the kit:
```ts
  const safeOp = await safeKit.createTransaction({ transactions });
  
  const safeOpHash = calculateSafeUserOperationHash(
    safeOp.data,
    // The following two values are part of the kit anyway!
    BigInt(await safeKit.getChainId()),
    await safeKit.protocolKit.getFallbackHandler(),
  );
```
If this lands, the code would look more like this:

```ts
  const safeOp = await safeKit.createTransaction({ transactions });
  const safeOpHash = await safeKit.hashSafeOperation(safeOp);
```

Note that this has was computed in the class already (which now uses the utility function).

### Alternative options:
1. It would actually be possible to put this method deeper into the `protocolKit` since both chainId and the `fallBackHandler` can be acquired from there.
2. We could also define the method on `SafeUserOperation` instead of `EthSafeOperation` (since only the data field is used to compute this).

### Test Plan

A unit test is added with expected hash.

### DIFF
--- a/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.test.ts
+++ b/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.test.ts
@@ -746,7 +746,6 @@ describe('Safe4337Pack', () => {
     })
 
     const safeOpHash = await safe4337Pack.hashSafeOperation(safeOperation)
-
-    expect(safeOpHash).toContain('ProvideExpectedHash')
+    expect(safeOpHash).toBe('0xb80f1998d7f72a3ee4c27d2db6b308d3744d15ad84d49840978f0d8fd39d82ab')
   })
 })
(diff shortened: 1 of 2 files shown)


bh2smith:

dasanra:
Hey @bh2smith we have just published `@safe-global/relay-kit` `v3.0.3`.

Instead of adding a new util to the Safe4337Pack class we added a `getHash` method in the SafeOperation.
https://github.com/safe-global/safe-core-sdk/blob/ea0d5018a93f294dfd891e6c8963edcb96431876/packages/relay-kit/src/packs/safe-4337/SafeOperation.ts#L94-L96

Could you check it that also works for you?

PR #890, opened 2024-07-01
feat: add short name for berachainbArtio

leonimella:
## What it solves
Adds short name for Berachain bArtio
https://chainlist.org/chain/80084

https://github.com/safe-global/safe-deployments/pull/676


dasanra:
Solved in #936

PR #892, opened 2024-07-03
add bob short name

DenSmolonski:
## What it solves
Extend short name for bob network


dasanra:
This chain is not part of https://github.com/safe-global/safe-deployments so the shortname won't be added

PR #905, opened 2024-07-09
feat(protocol-kit): Add deployment functionality to Safe class

### LINKED ISSUE #446: Add Safe deployment functionality to Safe class
## Context / issue

## Proposed solution

## Alternatives

DaniSomoza:
## What it solves
Resolves #446 

### Open Questions

1. Callback Parameter Removal: 
   - I have opted to remove the [`callback`](https://github.com/safe-global/safe-core-sdk/blob/main/packages/protocol-kit/src/SafeFactory.ts#L127) parameter from the new deploy() function that was present in the old `SafeFactory` class. We can reintroduce it if the team prefers.

2. Function Naming:
   - I chose to name the function `deploy()` instead of `deployAndSend()`. I am open to changing it if there are any objections or suggestions for a better name.

3. Return Type of `deploy()`:
   - Currently, the new `deploy()` method returns the new instance of the `protocol-kit` with the Safe already deployed. However, we can change this to return the transaction hash (`tx-hash`) or another value if that is preferable.
   
Please review and provide your feedback on these points.


packages/protocol-kit/tests/e2e/utilsContracts.test.ts
```diff
@@ -13,27 +13,34 @@ import {
...
 // test util funcion to deploy a safe (needed to check the expected Safe Address)
 async function deploySafe(
-  deploySafeProps: DeploySafeProps,
+  predictedSafeProps: PredictedSafeProps,
   provider: Eip1193Provider,
   contractNetworks: ContractNetworksConfig,
   signerAddress?: string
 ): Promise<Safe> {
-  const safeFactory = await SafeFactory.init({
+  const safeSDK = await Safe.init({
```
yagopv, line 30: safeSDK --> protocolKit

playground/protocol-kit/deploy-safe.ts
```diff
@@ -15,7 +15,7 @@ interface Config {
 }
 
 const config: Config = {
-  RPC_URL: 'https://sepolia.gateway.tenderly.co',
+  RPC_URL: 'https://rpc.sepolia.org',
```
yagopv, line 18: Why changed the RPC? If having issues with the tenderly one we should change it across the repo

Issue #869, opened 2024-06-17
[SDK Refactoring] `safe-kit` testing


yagopv:
## Context / issue

We want to create a new package offering an opinionated formula and a user-friendly experience for developers. We would like to offer an experience where we implement a top-level function ﻿`createSafeClient(options)` to interact with Safe. This new package will use the other kits internally, and it should be the primary package that most developers should use, leaving the other kits for more advanced implementations or core features.

Using the package should be similar to this code snippet:

```tsx
import { createSafeClient } from '@safe-global/safe-kit'

const safeClient = await createSafeClient({ signer, provider, chain })

const safeTxHash = await safeClient.send({ transactions })
```

The `send()` method will wrap the typical operations for `creatingTransaction`, `signTransaction` and `executeTransaction` making the developer experience better. The transaction signing would take place based on the provided `signer` 

```mermaid
flowchart LR
    A[send] -->|transaction| B(createTransaction)
    B(createTransaction) --> C(signTransaction)
    C(signTransaction) --> D(executeTransaction)
    A[send] -->|message| E(createMessage)
    E(createMessage) --> F(signMessage)
    F(signMessage) --> |if on-chain| D(executeTransaction)
    A[send] -->|userOperation| H(createRelayKitTransaction)
    H(createRelayKitTransaction) --> I(signSafeOperation)
    I(signSafeOperation) --> J(executeRelayKitTransaction)
```

## Proposed solution
We should create tests for this new package

## Additional context
[Research document](https://www.notion.so/safe-global/SDK-Refactor-4e45fd5f454d4715a60e72d4854b9aa9)

Related tickets
---

[Create Package](https://github.com/safe-global/safe-core-sdk/issues/866)
[Implement createSafeClient and send](https://github.com/safe-global/safe-core-sdk/issues/867)
[Add Multisig](https://github.com/safe-global/safe-core-sdk/issues/868)
[Add testing](https://github.com/safe-global/safe-core-sdk/issues/869)

PR #911, opened 2024-07-16
feat: Safe kit testing

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/869

## How this PR fixes it
This PR add unit testing to `safe-kit`


packages/safe-kit/src/SafeClient.ts
```diff
@@ -38,15 +41,25 @@ export class SafeClient {
...
   /**
    * Sends transactions through the Safe protocol.
+   * You can send an array to transactions { to, value, data} that we will convert to a transaction batch
    *
-   * @param {TransactionBase[]} transactions An array of transactions to be sent.
-   * @param {TransactionOptions} [options] Optional transaction options.
+   * @param {SendTransactionProps} props The SendTransactionProps object.
+   * @param {TransactionBase[]} props.transactions An array of transactions to be sent.
+   * @param {string} props.transactions.to The recipient address of the transaction.
+   * @param {string} props.transactions[].value The value of the transaction.
```
dasanra, line 49:
```suggestion
   * @param {string} props.transactions[].to The recipient address of the transaction.
   * @param {string} props.transactions[].value The value of the transaction.
```

packages/safe-kit/src/types.ts
```diff
@@ -1,6 +1,50 @@
...
+  validUntil?: number
+  validAfter?: number
+  feeEstimator?: IFeeEstimator
+}
+
+export type ConfirmSafeOperationProps = {
+  safeOperationHash: string
+}
+
+export type PaginationOptions = {
```
dasanra, line 43: Should this be the same type used in the api-kit? As it's going to be used to call the api-kit

packages/safe-kit/src/extensions/messages/SafeMessageClient.ts
```diff
@@ -77,7 +83,7 @@ export class SafeMessageClient {
    * @param {GetSafeMessageListProps} [options] Optional query parameters for pagination
    * @returns {Promise<SafeMessageListResponse>} A list of pending messages
    */
-  async getPendingMessages(options?: GetSafeMessageListProps): Promise<SafeMessageListResponse> {
+  async getPendingMessages(options?: PaginationOptions): Promise<SafeMessageListResponse> {
```
dasanra, line 86: Should this function also get an object with optional parameters?

yagopv: apart from the pagination options ? I can only see a parameter to get a concrete message and in this case can be a explicit method. What parameters are you thinking in?

packages/safe-kit/src/types.ts
yagopv: yes, now that i removed the `order` it doesn't make sense the explicit type

yagopv: done

dasanra (approved):

PR #917, opened 2024-07-19
feat: add autonomys testnet shortname

DenSmolonski:
I would like to add short name for Autonomys Tesnet


dasanra:
Solved in #936 

This PR was not adding the shortName as in ethereum chains list
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-490000.json

PR #919, opened 2024-07-19
fix(api-kit): Fix e2e tests for multisig transactions

tmjssz:
## What it solves
Fixes the confirmTransaction test which is currently sometimes running into timeouts on the viem migration branch (#887), due to uncaught errors from `isValidSignature` requests.

## How this PR fixes it
The test was incorrect because it used the same transaction hash repeatedly, leading to assertions being made on an already confirmed transaction instead of a newly proposed one. By adding a timestamp to the transaction data, we ensure that each transaction hash is unique and that a new transaction is proposed for every test run.

Additionally, the test incorrectly called the `isValidSignature` function with a transaction hash. This function is intended only for messages, causing the contract call to fail consistently. However, the error was not properly caught; the function merely returns `false` when an error occurs. With the migration to viem, the `isValidSignature` contract call sometimes takes too long until failing, resulting in occasional test timeouts. By removing the incorrect call, we prevent those timeout errors.

### DIFF
--- a/packages/api-kit/tests/e2e/confirmTransaction.test.ts
+++ b/packages/api-kit/tests/e2e/confirmTransaction.test.ts
@@ -4,16 +4,20 @@ import Safe, {
   SigningMethod,
   buildContractSignature
 } from '@safe-global/protocol-kit'
-import { SafeTransactionDataPartial } from '@safe-global/safe-core-sdk-types'
+import {
+  SafeMultisigConfirmationResponse,
+  SafeTransactionDataPartial
+} from '@safe-global/safe-core-sdk-types'
 import SafeApiKit from '@safe-global/api-kit/index'
 import chai from 'chai'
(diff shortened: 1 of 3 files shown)


packages/api-kit/tests/e2e/confirmTransaction.test.ts
```diff
@@ -90,14 +95,38 @@ describe('proposeTransaction', () => {
       safeAddress
     })
 
-    const isValidSignature = await protocolKit.isValidSignature(txHash, [ethSig, signerSafeSig])
-    console.log('- isValidSignature(txHash, signature) = ', isValidSignature)
-    // chai.expect(isValidSignature).to.be.true
```
tmjssz, line 95: With this uncommented, the test would have failed consistently

dasanra (approved):

PR #928, opened 2024-07-25
feat: Extend SafeProviders externalClient type

tmjssz:
## What it solves
Addresses the feedback provided in [this comment](https://github.com/safe-global/safe-core-sdk/pull/887#discussion_r1688315629).

## How this PR fixes it

This PR enhances the `SafeProvider` by enabling it to support both `PublicClient` and `WalletClient` as external providers. 

### Key Changes:
- **Type Alias Introduction**: A new type alias `ExternalClient` has been introduced, defined as `PublicClient | ExternalSigner`. This type is now used for the `#externalProvider` property in `SafeProvider`.
- **Method Adjustments**: Due to the extended type, `viem` methods are now individually imported from `viem/actions` instead of being directly invoked on the client.

These changes ensure that `SafeProvider` can handle a broader range of external providers, enhancing its flexibility and functionality.


tmjssz:
No idea why the build in the pipeline is failing. There is no issue when running the build locally 🤔

dasanra (approved):

PR #939, opened 2024-08-09
chore: Rename `safe-kit` to `sdk-starter-kit`

dasanra:
## What it solves
Sets the final name for `safe-kit` to be `sdk-starter-kit`


tmjssz (approved):

packages/sdk-starter-kit/src/types.ts
```diff
@@ -46,22 +46,23 @@ export type SafeConfig = {
   saltNonce?: string
 }
 
-export type ExistingSafeKitConfig = {
+export type ExistingSdkStarterKitConfig = {
```
leonardotc, line 49: Thing is that, now that its renamed, this name became really "avant-garde".

yagopv: Agree, I changed it

PR #944, opened 2024-08-14
feat: add autonomys testnet shortname

DenSmolonski:
I would like to add short name for Autonomys Tesnet


DenSmolonski:
recheck

dasanra:
Added in #936

PR #948, opened 2024-08-19
feat(protocol-kit): Add deployment functionality to Safe class + Viem

### LINKED ISSUE #446: Add Safe deployment functionality to Safe class

leonardotc:
## What it solves
Resolves #446 

### Open Questions

1. Callback Parameter Removal: 
   - I have opted to remove the [`callback`](https://github.com/safe-global/safe-core-sdk/blob/main/packages/protocol-kit/src/SafeFactory.ts#L127) parameter from the new deploy() function that was present in the old `SafeFactory` class. We can reintroduce it if the team prefers.

2. Function Naming:
   - I chose to name the function `deploy()` instead of `deployAndSend()`. I am open to changing it if there are any objections or suggestions for a better name.

3. Return Type of `deploy()`:
   - Currently, the new `deploy()` method returns the new instance of the `protocol-kit` with the Safe already deployed. However, we can change this to return the transaction hash (`tx-hash`) or another value if that is preferable.
   
Please review and provide your feedback on these points.

## Previous PR reference
https://github.com/safe-global/safe-core-sdk/pull/905


guides/integrating-the-safe-core-sdk.md
```diff
@@ -128,10 +116,13 @@ const safeAccountConfig: SafeAccountConfig = {
   threshold: 2,
   // ... (optional params)
 }
-const protocolKit = await safeFactory.deploySafe({ safeAccountConfig })
+
+const predictedSafe = await Safe.init({ provider, signer, predictSafe, contractNetworks })
```
yagopv, line 120:
```suggestion
const protocolKit = await Safe.init({ provider, signer, predictSafe, contractNetworks })
```

playground/protocol-kit/deploy-safe.ts
```diff
@@ -30,40 +30,44 @@ async function main() {
...
   // Config of the deployed Safe
   const safeAccountConfig: SafeAccountConfig = {
     owners: config.DEPLOY_SAFE.OWNERS,
     threshold: config.DEPLOY_SAFE.THRESHOLD
   }
+
   const saltNonce = config.DEPLOY_SAFE.SALT_NONCE
 
+  // Create SDK instance
+  const safeSDK = await Safe.init({
```
yagopv, line 42:
```suggestion
  let protocolKit = await Safe.init({
```

playground/protocol-kit/deploy-safe.ts
```diff
@@ -30,40 +30,44 @@ async function main() {
...
 
   // Deploy Safe
-  const safe = await safeFactory.deploySafe({
-    safeAccountConfig,
-    saltNonce,
-    callback
-  })
+  const deployedSafeSDK = await safeSDK.deploy()
+
+  console.log('Deployed Safe address:', await deployedSafeSDK.getAddress())
```
yagopv, line 67:
```suggestion
  console.log('Deployed Safe address:', await protocolKit.getAddress())
```

playground/protocol-kit/deploy-safe.ts
```diff
@@ -30,40 +30,44 @@ async function main() {
...
 
-  console.log('Deployed Safe:', await safe.getAddress())
+  // check if its deployed
+  console.log('Safe Account deployed: ', await deployedSafeSDK.isSafeDeployed())
```
yagopv, line 70:
```suggestion
  console.log('Safe Account deployed: ', await protocolKit.isSafeDeployed())
```

packages/protocol-kit/src/Safe.ts
```diff
@@ -264,13 +268,6 @@ class Safe {
    */
   async getAddress(): Promise<string> {
     if (this.#predictedSafe) {
-      const safeVersion = await this.getContractVersion()
```
yagopv, line 267: This is removed in several places but not sure why

yagopv:
About the open questions:

1) 👍🏻 It's ok for me to remove the callback
2) 👍🏻 `deploy()` is much better than `deployAndSend()` imo
3) 👎🏻 In this case, I'm not completely sure. My initial thought is that it's more about reinstating it internally and returning the transaction hash

PR #955, opened 2024-08-23
Refactor: static network in JsonRpcProvider

katspaugh:
## What it solves
The `staticNetwork` in the JSON provider options reduces the number of RPC requests made to detect the current chain (as a Safe Account always exists on just one chain).

See https://docs.ethers.org/v6/api/providers/jsonrpc/#JsonRpcApiProviderOptions


katspaugh:
Tests pass locally but on CI they failed. Probably flakiness?

yagopv (approved):

Issue #943, opened 2024-08-14
[SDK Refactor] New top level package `testing-kit`


dasanra:
## Context / issue
We have code repetition for testing. Between the `protocol-kit`, `api-kit`, and `relay-kit`, we share some common utilities that can be extracted into a shared package

## Proposed solution
Create a new `testing-kit` package. This package should include:

- `hardhat` configuration
- Contract deployments
- Common utils

This package allows us to remove code from the repository and concentrate the testing utilities in one place. This approach helps avoid errors and makes it easier to kickstart new packages.

## Additional context
[Research document](https://www.notion.so/safe-global/SDK-Refactor-4e45fd5f454d4715a60e72d4854b9aa9)

PR #962, opened 2024-08-27
feat: Testing kit

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/943

## How this PR fixes it
We are adding a  new package `testing-kit`
- Moved the hardhat configuration, contracts and helpers to the common package
- Generating a bin script for execute the tests in the pipeline or on local (Currently not working in the pipes bacause its not being generated in the target package .bin directory. Works locally. Perhaps because the package is not published?)
- Moved the scripts related to the `setupTests` function to the testing-kit
- Update protocol-kit and api-kit to use the new package. The hardhat configuration and deps are being removed from the target packages


DaniSomoza (approved):

package.json
```diff
@@ -35,7 +35,8 @@
     "ts-jest": "^29.1.5",
     "ts-node": "^10.9.2",
     "tsc-alias": "^1.8.8",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "@safe-global/testing-kit": "^1.0.0"
```
dasanra, line 39: It should not be necessary here as is set in the package specific ones

```suggestion
    "typescript": "^5.3.3"
```

packages/testing-kit/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "Smart Account",
+    "SDK",
+    "Testing"
+  ],
+  "scripts": {
+    "format:check": "prettier --check \"*/**/*.{js,json,md,ts}\"",
+    "format": "prettier --write \"*/**/*.{js,json,md,ts}\"",
+    "unbuild": "rimraf dist artifacts deployments cache .nyc_output *.tsbuildinfo",
+    "build": "yarn unbuild && hardhat compile && NODE_OPTIONS=--max-old-space-size=8192 tsc -p tsconfig.build.json && tsc-alias -p tsconfig.build.json && yarn copy-files && chmod +x dist/bin/testing-kit.js",
+    "copy-files": "cpx 'artifacts/**/*' dist/artifacts",
```
dasanra, line 20: Why do you need to copy them to this route?

packages/protocol-kit/tests/e2e/contractManager.test.ts
```diff
@@ -113,14 +97,14 @@ describe('Safe contracts manager', () => {
           Safe.init({
             provider,
             safeAddress,
-            contractNetworks: customContractNetworks
+            contractNetworks: customContractNetworks as ContractNetworksConfig
```
dasanra, line 100: Why do you prefer the cast? If you leave it at definition level will already assist you with the types if you need to modify the object.

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
```diff
@@ -352,7 +324,9 @@ describe('Off-chain signatures', () => {
         trusted: true,
         signatures: '0x111111222222'
       }
-      const signedTx = await safeSdk.signTransaction(safeServiceTransaction)
+      const signedTx = await safeSdk.signTransaction(
+        safeServiceTransaction as SafeMultisigTransactionResponse
```
dasanra, line 328: In this one is even more necessary to have the assistance at object creation time instead of casting.

packages/testing-kit/package.json
yagopv: A thing of the initial iterations that remained there. Good catch

packages/protocol-kit/tests/e2e/offChainSignatures.test.ts
yagopv: yes, changed

packages/protocol-kit/tests/e2e/contractManager.test.ts
yagopv: true, quick change 🙃

PR #931, opened 2024-08-02
feat: Viem migration + Passkeys

leonardotc:
## What it solves
Resolves issues #870 and #871.

## How this PR fixes it
This pull request migrates the entire repository to `viem`, removing `ethers` as a dependency. `ethers` is retained only as a devDependency for the test-utils in `api-kit` and `protocol-kit`.

### protocol-kit

* **SafeProvider**
  * `getExternalProvider` now returns a viem `PublicClient` (previously an ethers `Provider`).
  * `getExternalSigner` now returns a viem `WalletClient` (previously an ethers `AbstractSigner`).
  * `getTransaction` now returns a viem `Transaction` (previously an ethers `TransactionResponse`).
  * `encodeParameters` and `decodeParameters` now expect a `string` as the first parameter, containing the Human-readable ABI parameters (previously a `string[]`).
* All ABI parameters are now a viem `Abi` (previously `JsonFragment | JsonFragment[]`). This affects:
  * `defaultAbi` and `customContractAbi` in all Contract classes.
  * `customContractAbi` parameter in all `get...ContractInstance` functions and in the `GetContractProps` type.
* `getSafeProxyFactoryContractInstance` function's `signerOrProvider` parameter is now a viem `PublicClient`.

### api-kit

* `SafeApiKit`'s `addSafeDelegate` and `removeSafeDelegate` functions now expect a viem `WalletClient` as the signer prop (previously an ethers `Signer`).

### relay-kit

* `Safe4337Options` now expects a `bundlerClient` prop of viem's type `PublicClient` (previously ethers `JsonRpcProvider`).

### auth-kit

* `AuthKitEthereumProvider` type is now a viem `EIP1193Provider` (previously ethers `Eip1193Provider`). This new type restricts the RPC methods for EIP-1193, as opposed to the ethers type, which accepted any string as a method name.

## Other context

Replaces https://github.com/safe-global/safe-core-sdk/pull/887

PR #966, opened 2024-09-02
feat: add short name for bob-sepolia

leonimella:
## What it solves
Resolves: https://github.com/safe-global/safe-deployments/pull/761

## How this PR fixes it
Adds the short name for the BOB Sepolia network


dasanra:
Added in #931

PR #980, opened 2024-09-19
feat(protocol-kit): Add deployment functionality to Safe class

### LINKED ISSUE #446: Add Safe deployment functionality to Safe class

DaniSomoza:
## What it solves
Resolves #446

## How this PR fixes it

## Previous PR reference
#905
#948


guides/integrating-the-safe-core-sdk.md
```diff
@@ -102,36 +99,74 @@ const contractNetworks: ContractNetworksConfig = {
...
-  // ... (optional params)
+  owners: ['0x...', '0x...', '0x...'],
+  threshold: 2
+  // Additional optional parameters can be included here
 }
-const protocolKit = await safeFactory.deploySafe({ safeAccountConfig })
+
+const predictSafe = {
+  safeAccountConfig,
+  safeDeploymentConfig: {
```
dasanra, line 122: Is this still true?

dasanra (approved):

PR #992, opened 2024-09-30
chore: fix release comments

dasanra:
## What it solves
Fix some leftovers in the release and push the final versions


germartinez:
Question:
Can `import { SafeClient } from '@safe-global/sdk-starter-kit/SafeClient'` be replaced with `import { SafeClient } from '@safe-global/sdk-starter-kit` as it is exported [here](https://github.com/safe-global/safe-core-sdk/blob/main/packages/sdk-starter-kit/src/index.ts#L81)?

PR #995, opened 2024-10-03
feat: added Shape Sepolia Testnet

leonimella:
## What it solves
Resolves: https://github.com/safe-global/safe-deployments/pull/810

## How this PR fixes it
Adds the short name for the Shape Sepolia Testnet


dasanra:
Solved in #1017

PR #996, opened 2024-10-03
feat: added Shape Mainnet

leonimella:
## What it solves
Resolves: https://github.com/safe-global/safe-deployments/pull/811

## How this PR fixes it
Adds the short name for the Shape Mainnet


dasanra:
Solved in #1017

PR #1000, opened 2024-10-07
feat: add vana tesnet

DenSmolonski:
## What it solves
I've added vana moksha testnet EIP-3770 config

Link: https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-14800.json


dasanra:
Solved in #1017

PR #1022, opened 2024-10-23
chore: build protocol-kit with cjs and esm targets

frontendphil:
Over at Zodiac Pilot, we've had some trouble with the `protocol-kit` package, as it ships only as CJS. We're upgrading our codebase to the ESM future. 

This PR splits up the build for the `protocol-kit` package to create ESM and CJS sources. I've also extended the `package.json` to conform to everything from bundlers to Node JS native module resolution. 

You'll notice that I've removed a call to `tsc-alias`. AFAICT, the sources are bundled, which means any aliases have already been resolved. Please correct me if this assumption is wrong. I've tested this locally with our extension, and it works like a charm. We're now picking up the ESM build, and everything works as expected.

Let me know if anything is missing, if you need further clarification, or if anything else comes up!

### DIFF
--- a/packages/protocol-kit/package.json
+++ b/packages/protocol-kit/package.json
@@ -2,8 +2,14 @@
   "name": "@safe-global/protocol-kit",
   "version": "5.0.3",
   "description": "SDK that facilitates the interaction with Safe Smart Accounts",
-  "main": "dist/src/index.js",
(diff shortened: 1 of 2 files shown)


frontendphil:

dasanra:
We will be testing this solution, also the other kits should be compatible with esm and cjs

PR #1023, opened 2024-10-23
chore: build kits with cjs and esm targets

dasanra:
## What it solves
Resolves #


frontendphil:
I see that something seems off with this PR. Would it help if I opened up another PR that not only updates the build for `protocol-kit` but also takes a swing at all the builds in the reop? Building through `tsc` is pretty memory-heavy and can be done much faster with `build.` So, I could replicate the setup I added to the `protocol-kit` and add it to all the packages. WDYT?

frontendphil:
I've opened up another PR that addresses this topic with an extended scope. Would be great if you could have a look: https://github.com/safe-global/safe-core-sdk/pull/1047

This PR is obsolete now.

PR #1029, opened 2024-10-28
feat: change chain 88 shortname "tomo" -> "vic"

vickyvickydev:
## What it solves
Resolves #
- Chain ID 88 (formerly Tomochain) has renamed to Viction.

## How this PR fixes it
- Change short name from "tomo" to "vic"


vickyvickydev:

PR #1032, opened 2024-10-31
feat: Add deploymentType

yagopv:
## What it solves
- Adding the `deploymentType` prop in all the kits
- Added `safeVersion` as deployment option in the `sdk-starter-kit`
- Use `canonical` as default in the `sdk-starter-kit `and fallback to choose the first deployment in the list
- Add `txServiceUrl` config option to `sdk-starter-kit` for chains without the service


packages/protocol-kit/src/managers/contractManager.ts
```diff
@@ -55,20 +55,23 @@ class ContractManager {
         safeVersion,
         isL1SafeSingleton,
         customSafeAddress: safeAddress,
-        customContracts
+        customContracts,
+        deploymentType: predictedSafe?.safeDeploymentConfig?.deploymentType
```
dasanra, line 59: In this `else` this is never possible for any of the contracts. The `if` clause checks if predictedSafe option exists in the config

packages/sdk-starter-kit/src/types.ts
```diff
@@ -44,6 +44,7 @@ export type SafeConfig = {
   owners: string[]
   threshold: number
   saltNonce?: string
+  deploymentType?: DeploymentType
```
dasanra, line 47: This will suggest that when Predicting a Safe config the developer can set the deploymentType which is not true

dasanra (approved):

Issue #1026, opened 2024-10-25
[React Native] Add passkey compatibility for iOS


dasanra:
## Context / issue

Add passkey compatibility for iOS

## Proposed solution

Prepare the code for using Passkeys in iOS. Currently the protocol-kit needs to be refactored to allow passkeys not only for browsers.

During this implementation necessary changes for the protocol kit to be system agnostic needs to be done. This will already be compatible with Android system too.

Issue #1027, opened 2024-10-25
[React Native] Passkey compatibility for Android devices


dasanra:
## Context / issue

Add compatibility to the protocol-kit for Passkeys created by Android devices.

## Proposed solution

Create the code example and materials showing how to use Passkeys in Android devices when using the protocol-kit. Protocol kit should be adapted first in #1026 to be device agnostic.

PR #1033, opened 2024-10-31
feat(protocol-kit): Add react native compatibility

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/1026, https://github.com/safe-global/safe-core-sdk/issues/1027

## How this PR fixes it
- Creating a new static method to create the passkey signer
```
const credential = window.navigator.credentials.create(requestJson)
const passkeySigner = await createPasskeySigner(credential)
const protocolKit = await Safe.init({ provider, signer: passkeySigner })
```
- Adding a `getFn` to the `PasskeyArgType` allowing the credential retrieval
```
type GetPasskeyCredentialFn = (
  options?: PasskeyCredentialRequestOptions
) => Promise<PasskeyCredential<PasskeyAuthenticatorAssertionResponse>>
```
- Implementing a specific decoder for public keys coming from React Native


packages/protocol-kit/src/Safe.ts
```diff
@@ -1698,6 +1700,15 @@ class Safe {
...
   }
+
+  /**
+   * This method creates a signer to be used with the init method
+   * @param {Credential} credential - The credential to be used to create the signer. Can be generated in the web with navigator.credentials.create
+   * @returns {PasskeyArgType} - The signer to be used with the init method
+   */
+  static createPasskeySigner = async (credential: Credential): Promise<PasskeyArgType> => {
+    return extractPasskeyData(credential)
+  }
```
DaniSomoza, line 1711: we should also update our documentation in safe-docs, to use `createPasskeySigner` instead of `extractPasskeyData`

DaniSomoza (approved):

packages/protocol-kit/src/Safe.ts
yagopv: I already created a ticket for that

PR #1034, opened 2024-10-31
feat: add morph mainnet shortname

leonimella:
## What it solves
Adds the short name of Morph Mainnet network


dasanra:
Done in #1039

PR #1038, opened 2024-11-04
feat: add shortname for ink sepolia

leonimella:
## What it solves
Adds shortname for Ink Sepolia network


dasanra:
Done in #1067

PR #1041, opened 2024-11-05
feat(relay-kit): Control errors in `eth_estimateUserOperationGas` calls

yagopv:
## What it solves

Errors in `eth_estimateUserOperationGas` can happen easily so we are logging and throwing an exception on it


dasanra:
@yagopv a test fails because of this change

yagopv:
> @yagopv a test fails because of this change

Yes :( . I'm throwing the error instead the hardcoded message and now the test should work

PR #1042, opened 2024-11-06
feat: add story odyssey shortname

leonimella:
## What it solves
Adds shortname for Story Odyssey network


dasanra:
Done in #1067

PR #1047, opened 2024-11-13
chore: update build for the whole repo

frontendphil:
## What it solves
This PR updates the build for the whole repo. 

## How this PR fixes it

- All packages produce a CJS and ESM build
- All packages specify their exports according to what bundlers and Node need to understand it
- Tests of different packages do not require the internals of other packages
  - added new `package/test-utils` entry point where necessary
- The TS setup has been updated to address an OOM error that was caused by the last PR
  - path aliases are specified in the projects that need them instead of inside the global config
  - the `baseUrl` property is set on the package level, which reduces the memory footprint of a `tsc` build

## What is still open

Some `api-kit` tests started failing. AFAICT, this has something to do with the `PRIVATE_KEY` environment variable not being set to the correct value. However, I wonder why this wasn't an issue earlier, and I also can't pinpoint a change I made that would result in this variable not being picked up correctly. I would appreciate help looking into this, as I think I've addressed all other issues. 

The build for `relay-kit` is also a bit weird because the tests in other packages needed to require deeply into this package to mock internals. That's not great, but to "properly" fix this, I think the general test setup needs to be adjusted, and that's definitely out of the scope of what I'm trying to achieve here. 

There is also one test in `protocol-kit` that now times out. But AFAICT it only does that on CI. When I run the tests locally, they all pass.


dasanra:
@frontendphil Hey thanks for such work. Looks very promising.

Yes don't worry about the remaining CI, secrets are not accessible from forks. We will do a review and bring the code to an integration branch to completely test CI.

PR #1049, opened 2024-11-15
chore: build kits with cjs and esm targets

dasanra:
## What it solves
Continues #1047


frontendphil:
Hey folks! I wanted to give this a little bump because I'm working on the next thing, and once again, I'm seeing your code in a stack trace related to the whole CJS/ESM interop stuff. As always, If there is anything I can do to help push this forward, just let me know.

DaniSomoza:
Hi @frontendphil 

We’ve made significant progress and, after running multiple tests, it seems to be working well overall.

However, there’s one issue related to CommonJS compatibility that we’re still trying to address.

Specifically, due to the way the Safe class from the protocol-kit is exported, users currently need to use .default when importing it in CommonJS:

```javascript
const Safe = require('@safe-global/protocol-kit').default // TODO: remove .default

Safe.init({ key: 'value' })
```

We’d love to eliminate the need for `.default` to simplify integration for CommonJS, but we haven’t found a clean, simple solution without introducing extra complexity.

Thanks for your time and support.

frontendphil:
Hey @DaniSomoza 

Thanks for getting back to me. This probably won't be the answer you're looking for, but I need to ask anyway :) Would eliminating the `default` export (which [is](https://basarat.gitbook.io/typescript/main-1/defaultisbad) [discouraged](https://webdeveloper.beehiiv.com/p/avoid-default-exports-javascript-modules) [anyway](https://medium.com/@scr2em/please-avoid-default-export-f3b669668f33)) and replacing it with a named export be something you'd consider? 

Then, instead of:
```ts
const Safe = require('@safe-global/protocol-kit').default
```

People would need to do:
```ts
const { Safe } = require('@safe-global/protocol-kit')
```

This would mean a breaking API change, but it may also be worth considering.

DaniSomoza:
The changes are now ready for review.

I've published the following packages as `alpha` version for testing purposes:
- @safe-global/api-kit → `2.5.8-alpha.0`
- @safe-global/protocol-kit → `5.2.1-alpha.0`
- @safe-global/relay-kit → `3.4.1-alpha.1`
- @safe-global/sdk-starter-kit → `1.1.3-alpha.0`
- @safe-global/types-kit → `1.0.2-alpha.0`
- @safe-global/testing-kit → `0.0.1-alpha.0`


I've tested these versions in several of our demo projects, and everything seems to be working as expected. 🚀

CC: @frontendphil 

Thanks!

DaniSomoza (approved):

Issue #1054, opened 2024-11-21
Passkey signer error when using Bitwarden extension


nlordell:
> This issue was originally posted by @richardhenry in https://github.com/safe-global/safe-modules/issues/500

## Description

One of our users is experiencing issues using the passkey signer with their Safe wallet in my dapp. I think the issue is that they are using the Bitwarden Chrome extension for passkeys/passwords, which doesn't seem to provide passkey data as an ArrayBuffer. This is probably non-standard behavior, but Bitwarden otherwise works fine for general passkey sign-in, so I suspect that maybe the WebAuthn code in the Safe library needs to be more robust.

Here is the error that they were seeing in their browser console:

![Screenshot_2024-11-18_at_6 59 35_PM](https://github.com/user-attachments/assets/c855c026-e88b-4c48-8c32-873ab78c3d7d)

## Environment

- @safe-global/protocol-kit@5.0.2
- @safe-global/relay-kit@3.2.4
- Chrome 130.0.6723.91 (Official Build) (arm64)
- macOS Sonoma 14.6
- Bitwarden browser extension 48.0

## Steps to reproduce

1. Use Google Chrome with the Bitwarden browser extension
2. Call signSafeOperation with a passkey signer

richardhenry:
It looks like Bitwarden is providing a Uint8Array instead of an ArrayBuffer, and I'm also seeing just a plain Array type in our logs from some other unknown password manager. I have a fix for this that is working for our users - I'll open a PR. Thanks!

PR #1094, opened 2025-01-08
fix(protocol-kit): Fix non-ArrayBuffer passkey signature error (#1054)

richardhenry:
Some non-standards compliant password managers (including Bitwarden) provide the passkey signature as a Uint8Array or as an Array of numbers instead of an ArrayBuffer.

This causes a "First argument to DataView constructor must be an ArrayBuffer" error in extractSignature.

This change makes this code robust to such password managers.

Solves #1054


richardhenry:

richardhenry:
recheck

dasanra (approved):

Issue #1088, opened 2024-12-23
[APIKit] Type mistmatch between txs-service and SDK


gjeanmart:
## Description

There is a mismatch between the APIKit result and the transaction-service response, leading to an `undefined` value.

The transaction-service returns a `masterCopy` field as part of the `/api/v1/safes/<safe>/` endpoint response:

<img width="969" alt="image" src="https://github.com/user-attachments/assets/3aabc37a-6c54-4c3d-90a3-d72f7af2b347" />

However, APIKit expects and handles this value under the `singleton` field:

https://github.com/safe-global/safe-core-sdk/blob/main/packages/api-kit/src/types/safeTransactionServiceTypes.ts#L44-L54
```typescript
export type SafeInfoResponse = {
  readonly address: string
  readonly nonce: number
  readonly threshold: number
  readonly owners: string[]
  readonly singleton: string // <<<<< HERE
  readonly modules: string[]
  readonly fallbackHandler: string
  readonly guard: string
  readonly version: string
}
```

This discrepancy results in `safe.singleton` being `undefined` when calling:

```typescript
const safe = await apiKit.getSafeInfo(safe)
```

## Environment
- Safe Core SDK versions:
  
  - api-kit: 2.5.4
  - protocol-kit:
  - relay-kit:
  - types-kit:
  - sdk-starter-kit:
- Safe contract version: 
- Environment:
  - browser
  - non-browser

## Steps to reproduce

```typescript
const safe = '0x...' // some safe on Mainnet
const apiKit = new SafeApiKit({
  chainId: 1
})
const safe = await apiKit.getSafeInfo(safe)
expect(safe.singleton).toBeDefined() // fails
```

## Expected result

The response should consistently return a `singleton` field, correctly mapping the `masterCopy` value from the transaction-service response.

## Additional context

**Workaround:** 

```typescript
const safe = '0x...' // some safe on Mainnet
const apiKit = new SafeApiKit({
  chainId: 1
})
const safe = await apiKit.getSafeInfo(safe)
expect(safe['masterCopy']).toBeDefined() // success
```

Issue #1085, opened 2024-12-18
Get Incoming Transaction response type


salochara:
Hi!

I am using the [.getIncomingTransactions()](https://github.com/safe-global/safe-core-sdk/blob/1cff4b5ed70efc58dc6699e140cee0774c13cb83/packages/api-kit/src/SafeApiKit.ts#L486) from the SafeApiKit using a Safe that was deployed with Safe4437. 

Whenever I call this method, I get the `tokenInfo` property which is not defined in the `TransferListResponse` type. 

TypeScript correctly complains about it, but I am actually getting that `tokeInfo` property populated whenever I am dealing with ERC-20 tokens.
And also I am getting the `transferId`  property that's also not defined in `TransferListResponse` type

```
  incoming transactions {
  count: 3,
  next: null,
  previous: null,
  results: [
  {
      type: 'ERC20_TRANSFER',
      executionDate: '...',
      blockNumber: ...,
      transactionHash: '....',
      to: '...',
      value: '...',
      tokenId: null,
      tokenAddress: '...',
      transferId: '...',
      tokenInfo: [Object],
      from: 'Ox'
   },
  ]}
  ```

PR #1100, opened 2025-01-16
fix(api-kit): add missing properties to types

### LINKED ISSUE #1064: SafeMultisigTransactionResponse dataDecoded field's type
## Description

Hi there,

I’ve recently started using the safe-sdk and came across what seems to be a type mismatch. When using the `getMultisigTransactions` method from SafeApiKit (imported from @safe-global/api-kit) on an instance of new SafeApiKit, the API returns structured data for the `dataDecoded` field. However, the `SafeMultisigTransactionResponse` type, which is the return type of the `getMultisigTransactions` method, defines `dataDecoded as a string` as seen [here](https://github.com/safe-global/safe-core-sdk/blob/014f365e46cf278897c5ece45c1490c7c749f2d7/packages/types-kit/src/types.ts#L234).

Could you confirm if this is expected behavior or if the type definition needs to be updated?

Thanks for your help!

What I do is very simple:

```typescript

const getTransactionList = async () => {
    const safeApiClient = new SafeApiKit({ chainId }) // I put the instance creation here for simplicity of having the relevant line of codes
    const trxList = await safeApiClient.getMultisigTransactions(
      process.env.NEXT_PUBLIC_SAFE_WALLET_ADDRESS
    )
    
   return trxList
}
```

What it returns:

Screenshot 2024-12-02 at 18 59 32

## Environment
- Safe Core SDK versions:
  
  - api-kit: `2.5.4`
  - protocol-kit: `5.0.4`
  - relay-kit: `3.2.4`
  - types-kit: `1.0.0`
  - sdk-starter-kit:
- Safe contract version: `1.4.1+L2`
- Environment:
  - browser: chrome

## Expected result

The field of `decodedData` [here](https://github.com/safe-global/safe-core-sdk/blob/014f365e46cf278897c5ece45c1490c7c749f2d7/packages/types-kit/src/types.ts#L234) respects the return type of the api field.

## Additional context

I'm using the sdk in a nextjs project.

dasanra:
## What it solves
Resolves #1064, #1085, #1088

## How this PR fixes it
Adds missing properties to types


packages/api-kit/src/SafeApiKit.ts
```diff
@@ -258,6 +258,14 @@ class SafeApiKit {
     return sendRequest({
       url: `${this.#txServiceBaseUrl}/v1/safes/${address}/`,
       method: HttpMethod.Get
+    }).then((response: any) => {
+      // FIXME remove when service returns the singleton property
```
DaniSomoza, line 262:
```suggestion
      // FIXME remove when the service always returns the singleton property
```

I Think that this small change makes clear that the condition for removal is when the API always consistently returns the singleton property.

DaniSomoza (approved):

Issue #1101, opened 2025-01-17
[4337] Support 0.7 Entrypoint implementation e2e testing


yagopv:
## Context / issue

0.7 entrypoint was unfinished when we started 4337 implementation. Currently it seems ready to be used and it should be possible to use it with the Safe SDK. We need to test and support both versions in the SDK.

## Proposed solution

Review the playgrounds and test the following use cases for both the 0.6 and 0.7 entry point versions

- [x] Token transfer
- [x] Token transfer with counterfactual deployment
- [x] Token transfer with ERC20 paymaster
- [x] Token transfer with ERC20 paymaster and counterfactual deployment
- [x] Token transfer sponsored transactions
- [x] Token transfer sponsored transactions with counterfactual deployment

PR #1103, opened 2025-01-17
feat(relay-kit): Add Entrypoint v0.7 support

### LINKED ISSUE #845: [4337] Support 0.7 Entrypoint implementation

yagopv:
## What it solves
Resolves #845 https://github.com/safe-global/safe-core-sdk/issues/1101

## How this PR fixes it
We are adding support for entrypoint v0.7

Now you can choose `safeModulesVersion` 0.3.0 when calling the init() method of the `Safe4337Pack`. This module version is the one compatible with entrypoint v0.7

#### Upgrade path
For existing Safes that want to upgrade to use the v0.7 version:
1) Should enable the 0.3.0 module
2) Should update the fallback handler to be the new enabled module

Could be through the protocol-kit or through the transaction service

### DIFF
--- a/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.test.ts
+++ b/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.test.ts
@@ -9,7 +9,7 @@ import {
 } from '@safe-global/safe-modules-deployments'
 import { MetaTransactionData, OperationType } from '@safe-global/types-kit'
 import { Safe4337Pack } from './Safe4337Pack'
-import SafeOperation from './SafeOperation'
+import BaseSafeOperation from './BaseSafeOperation'
 import * as constants from './constants'
 import * as utils from './utils'
 import {
@@ -382,7 +382,7 @@ describe('Safe4337Pack', () => {
         transactions
       })
 
-      expect(safeOperation).toBeInstanceOf(SafeOperation)
+      expect(safeOperation).toBeInstanceOf(BaseSafeOperation)
       expect(safeOperation.getSafeOperation()).toMatchObject({
         safe: fixtures.SAFE_ADDRESS_v1_4_1_WITH_0_3_0_MODULE,
         entryPoint: fixtures.ENTRYPOINT_ADDRESS_V07,
@@ -418,7 +418,7 @@ describe('Safe4337Pack', () => {
         transactions: [transferUSDC]
       })
 
-      expect(safeOperation).toBeInstanceOf(SafeOperation)
+      expect(safeOperation).toBeInstanceOf(BaseSafeOperation)
       expect(safeOperation.getSafeOperation()).toMatchObject({
         safe: fixtures.SAFE_ADDRESS_v1_4_1_WITH_0_3_0_MODULE,
         entryPoint: fixtures.ENTRYPOINT_ADDRESS_V07,
@@ -480,7 +480,7 @@ describe('Safe4337Pack', () => {
         transactions: [transferUSDC]
       })
 
-      expect(sponsoredSafeOperation).toBeInstanceOf(SafeOperation)
+      expect(sponsoredSafeOperation).toBeInstanceOf(BaseSafeOperation)
       expect(sponsoredSafeOperation.getSafeOperation()).toMatchObject({
         safe: fixtures.SAFE_ADDRESS_v1_4_1_WITH_0_3_0_MODULE,
         entryPoint: fixtures.ENTRYPOINT_ADDRESS_V07,
@@ -541,7 +541,7 @@ describe('Safe4337Pack', () => {
 
       const batch = [transferUSDC, approveTransaction]
 
-      expect(sponsoredSafeOperation).toBeInstanceOf(SafeOperation)
+      expect(sponsoredSafeOperation).toBeInstanceOf(BaseSafeOperation)
       expect(sponsoredSafeOperation.getSafeOperation()).toMatchObject({
         safe: fixtures.SAFE_ADDRESS_v1_4_1_WITH_0_3_0_MODULE,
         entryPoint: fixtures.ENTRYPOINT_ADDRESS_V07,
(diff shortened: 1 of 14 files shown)


playground/utils.ts
```diff
@@ -1,6 +1,16 @@
...
+  createPublicClient,
+  custom,
+  encodeFunctionData,
+  formatEther,
+  parseAbi
+} from 'viem'
 import { Safe4337Pack } from '@safe-global/relay-kit'
 import { ExternalSigner } from '@safe-global/protocol-kit'
+import { getBlock, waitForTransactionReceipt } from 'viem/actions'
+import { MetaTransactionData } from 'packages/types-kit/dist/src'
```
dasanra, line 12:
```suggestion
import { MetaTransactionData } from '@safe-global/types-kit'
```

packages/relay-kit/src/packs/safe-4337/types.ts
```diff
@@ -184,25 +177,35 @@ type UserOperationStringValues = Omit<
...
-    Parameters: [UserOperationStringValues, string, { sponsorshipPolicyId?: string }?]
-    ReturnType: {
-      paymasterAndData: string
-      callGasLimit: string
-      verificationGasLimit: string
-      preVerificationGas: string
-    }
+    Method: RPC_4337_CALLS.GET_PAYMASTER_DATA
+    Parameters: [UserOperationStringValues, string, string, { token: string }?]
+    ReturnType:
```
dasanra, line 198: These are Pimlico specific methods. Why are they defined in the general types instead of Pimlico estimator specific types?

packages/protocol-kit/src/utils/signatures/utils.ts
```diff
@@ -1,11 +1,15 @@
...
 import SafeProvider from '@safe-global/protocol-kit/SafeProvider'
-import { SafeSignature, SafeEIP712Args, SafeTransactionData } from '@safe-global/types-kit'
-import semverSatisfies from 'semver/functions/satisfies.js'
+import {
+  SafeSignature,
+  SafeEIP712Args,
+  SafeTransactionData,
+  SigningMethod
+} from '@safe-global/types-kit'
+import semverSatisfies from 'semver/functions/satisfies'
```
dasanra, line 9:
```suggestion
import semverSatisfies from 'semver/functions/satisfies.js'
```

`.js` was added to this import for dual builds to work.

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.test.ts
```diff
@@ -73,30 +76,33 @@ describe('Safe4337Pack', () => {
...
+          safeModulesVersion: '0.2.0'
         })
       ).rejects.toThrow(
-        'Incompatibility detected: The EIP-4337 fallbackhandler is not attached to the Safe Account. Attach this fallbackhandler (address: 0xa581c4A4DB7175302464fF3C06380BC3270b4037) to ensure compatibility.'
+        `Incompatibility detected: The EIP-4337 fallbackhandler is not attached to the Safe Account. Attach this fallbackhandler (address: ${fixtures.SAFE_4337_MODULE_ADDRESS_V0_2_0}) to ensure compatibility.`
       )
     })
 
-    it('should throw an error if the Safe Modules do not match the supported version', async () => {
+    it('should throw an error if the 4337 Module is not enabled in the Safe account', async () => {
```
dasanra, line 98: This title was changed but the test still checks the same
```suggestion
    it('should throw an error if the Safe Modules do not match the supported version', async () => {
```

packages/relay-kit/src/packs/safe-4337/SafeOperationBase.ts
```diff
@@ -0,0 +1,73 @@
...
+      domain: {
+        chainId: Number(this.options.chainId),
+        verifyingContract: this.options.moduleAddress
+      },
+      types: this.getEIP712Type(),
+      primaryType: 'SafeOp',
+      message: this.getSafeOperation()
+    })
+  }
+
```
dasanra, line 67: This two methods are missing in the unit tests

playground/utils.ts
```diff
@@ -56,3 +66,73 @@ export async function transfer(
...
+  }
+
+  // Fund Safe
+  if (nativeTokenAmount) {
+    console.log(`sending ${formatEther(nativeTokenAmount)} native tokens...`)
+
+    const hash = await externalSigner?.sendTransaction({
+      to: senderAddress,
+      value: nativeTokenAmount,
+      chain: sepolia
```
dasanra, line 109: This sets up 4337 playground but is not getting the chain from the .env-sample that is inside relay-kit folder. This may cause issues if someone intends to use a chain different than sepolia.

dasanra (approved):

PR #1104, opened 2025-01-17
feat (flow): add short names for Flow

DenSmolonski:
## What it solves
Add short names for Flow and Flow Testnet 

https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-747.json
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-545.json


DenSmolonski:
@dasanra Hi, I hope you are doing well! I just wondering if there are any updates?

dasanra:
@DenSmolonski before the chain is added to the SDK needs to be available in `@safe-global/safe-deployments`. This was done just 4 days ago, so the chain will be available with the next `safe-deployments` dependency push in the next release.

dasanra:
Will be added by #1151

PR #1114, opened 2025-01-29
fix(protocol-kit): Fix signature generation for multiple contract signatures

sidnym-ladrut:
## What it solves
There are currently a few minor issues in `protocol-kit` related to contract signatures (i.e. [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271)):

1. Contract signatures processed by the API are not recognized as such (fix in `Safe.ts`).
2. The dynamic part of a contract signature is calculated as the entire segment of the given data, when it should be the [last 97 bytes only](https://sepolia.etherscan.io/address/0x29fcb43b46531bca003ddc8fcb67ffe91900c762#code#F2#L286) (65 bytes signature + 32 bytes length).

## How this PR fixes it
These issues are fixed in the following ways:

1. All `CONTRACT_SIGNATURE` responses are registered as contract signatures.
2. The dynamic part of contract signatures is adjusted to be the last 97 bytes of the data payload.

### DIFF
--- a/packages/protocol-kit/src/Safe.ts
+++ b/packages/protocol-kit/src/Safe.ts
@@ -1268,7 +1268,11 @@ class Safe {
     })
     serviceTransactionResponse.confirmations?.map(
       (confirmation: SafeMultisigConfirmationResponse) => {
-        const signature = new EthSafeSignature(confirmation.owner, confirmation.signature)
+        const signature = new EthSafeSignature(
+          confirmation.owner,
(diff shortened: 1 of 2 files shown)


sidnym-ladrut:

Bullet1010 (approved):

Bullet1010 (approved):

Bullet1010:
need attention

PR #1124, opened 2025-02-06
Update pimlico endpoint in examples

mouseless0x:
## What it solves
Related to [safe-docs/pull/684](https://github.com/safe-global/safe-docs/pull/684)

## How this PR fixes it
Updates the pimlico api endpoint in examples to use the `add_balance_override` query param


mouseless0x:

valle-xyz:
Thanks @mouseless0x for opening this requested PR. LGTM.

yagopv:
Hey @valle-xyz , this change is not really necessary. We moved the URL to an [env variable](https://github.com/safe-global/safe-core-sdk/blob/development/playground/relay-kit/.env-sample)

Apart from that, I think if the new behavior is to not override balances we should stick to it

PR #1132, opened 2025-02-12
Exclude AddressType declaration from types build

### LINKED ISSUE #1087: Type Safety Issue: Global Module Augmentation Weakens Address Types
## Description

Installing `@safe-global/api-kit` changes the global type definition for Ethereum addresses from `0x${string}` to `string` through a module augmentation of `abitype`:

```typescript
// In @safe-global/api-kit/dist/index.d.ts
declare module 'abitype' {
  interface Register {
    AddressType: string;
  }
}
```
This global change affects all packages that depend on `abitype` for their types (including `viem` and by extension `wagmi`), weakening type safety across the entire consuming application.

### Impact
1. Changes `viem`'s `Address` type from `0x${string}` to `string` project-wide
2. Affects type safety in consuming applications that rely on strict address typing
3. Makes it harder to catch malformed addresses at compile time
4. Prevents applications from maintaining stricter address types even in their own code

### Attempted Solutions
1. Creating a local module augmentation to restore the stricter type results in TypeScript resolving the type to `any`
2. No way to "opt out" of the global type modification

## Environment
- Safe Core SDK versions:
  
  - api-kit: 2.5.4
- viem: 2.21.55
- TypeScript: 5.7.2

## Steps to reproduce
1. Go to https://www.typescriptlang.org/play
2. First import Address from `viem` alone and inspect the type
3. Import API Kit before `viem` and inspect type again

tsplayground-viem

image

## Expected result

Installing `@safe-global/api-kit` should not affect the type definitions of other packages. The `Address` type from viem should remain `0x${string}` regardless of whether Safe API Kit is installed or not.

DaniSomoza:
## What it solves
Resolves #1087 

This PR ensures that the `AddressType` override is only applied within our packages (without affecting external projects).

```typescript
declare module 'abitype' {
  interface Register {
    AddressType: string
  }
}
```

Previously, the `declare module 'abitype'` was applied globally, which unintentionally impacted other packages like `viem` and `wagmi`. This could lead to weakened type safety across projects that use our packages.

## How this PR fixes it

- Added `local-declarations.d.ts` to each package **without modifying `tsconfig.json`**.  
- Updated `tsconfig.build.json` to **exclude `local-declarations.d.ts` from the type build** Preventing it from being included in the final output.  

✅ `AddressType` is still overridden to `string` inside our packages.
✅ `viem` and other dependencies remain **unaffected** by this change for other projects that uses our packages.  
✅ TypeScript builds properly without generating unwanted declarations.  

Now our packages work as expected without introducing global type modifications.

### DIFF
--- a/packages/api-kit/src/index.ts
+++ b/packages/api-kit/src/index.ts
@@ -3,9 +3,3 @@ import SafeApiKit, { SafeApiKitConfig } from './SafeApiKit'
 export * from './types/safeTransactionServiceTypes'
 export { SafeApiKitConfig }
 export default SafeApiKit
-
-declare module 'abitype' {
-  export interface Register {
-    AddressType: string
-  }
-}
--- a/packages/api-kit/src/types/local-declarations.d.ts
+++ b/packages/api-kit/src/types/local-declarations.d.ts
@@ -1,0 +1,7 @@
+export {}
+
+declare module 'abitype' {
+  interface Register {
+    AddressType: string
+  }
+}
--- a/packages/api-kit/tsconfig.build.json
+++ b/packages/api-kit/tsconfig.build.json
@@ -7,5 +7,5 @@
     "outDir": "dist/src"
   },
   "include": ["src/**/*"],
-  "exclude": ["dist", "test"]
+  "exclude": ["dist", "test", "local-declarations.d.ts"]
 }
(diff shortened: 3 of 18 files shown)


twodam:
when can this be merged?

dasanra:
Closed as solution at #1167 is preferred

Issue #1152, opened 2025-02-26
eth_chainId request sent to paymaster url instead of bundler


itsdeka:
I assume you're developing the latest version of safe-core-sdk using pimlico.
Unfortunately this makes the library not compatible with many other providers.

PimlicoFeeEstimator at line 927 uses the paymasterClient to get the chain id. This fails using another provider (for example Candide).

A simple solution would be to retrieve the chain id using a bundlerClient

const bundlerClient = createBundlerClient(
     bundlerUrl
);
const chainId = await this.#getChainId(bundlerClient);

With this change only it's possible to use Pimlico as bundler and Candide (or any other provider) as paymaster.

Ideally there should be a generic fee estimator that does not use pimlico_getUserOperationGasPrice but retrieve maxFeePerGas and maxPriorityFeePerGas from any other source

dasanra:
The Kit was created in a way that can be used with any provider. Pimlico is provided as default but can be changed for any other custom implementation.

When using the `createTransaction` or `getEstimateFee` methods, one of the options is the `feeEstimator`. By default will rely in Pimlico provided one, but the developer can create their own, following the `IFeeEstimator` interface. Almost all providers have some custom methods for gasEstimation, so it was not possible for us to create a generic fee estimator.

A consideration that needs to be done is that we are about to release a new major version which will change the IFeeEstimator interface.

mholtzman:
Couldn't a generic estimator be written using more standard methods like `eth_estimateUserOperationGas`? I'm running into similar issues and I'm not really sure it's reasonable to expect developers to hand roll a fee estimator as a prerequisite to using this SDK.

itsdeka:
@mholtzman take a look https://github.com/safe-global/safe-core-sdk/pull/1153

yagopv:
Hey!, If you pass an empty estimator to the `createTransaction` method then only the `eth_estimateUserOperationGas` is [going to be called](https://github.com/safe-global/safe-core-sdk/blob/e5aac7776b7e17b62ea088529de7748104e2d080/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts#L407-L453)

```
createTransaction({ transactions, options: { feeEstimator: {} } })
```

Sednaoui:
[...]

PR #1153, opened 2025-02-26
Use bundlerUrl to retrieve chain id

itsdeka:
https://github.com/safe-global/safe-core-sdk/issues/1152


dasanra:
Solved by #1238

PR #1160, opened 2025-03-07
Fix: SafeERC4337Pack Compatibility with Entrypoint v0.7

Another-DevX:
## What it solves
The [Safe SDK documentation](https://docs.safe.global/sdk/relay-kit/guides/4337-safe-sdk) states:

> We have added support for the Entrypoint v0.7 contract but we are not making it the default yet. If you are using Entrypoint v0.7, you need to set the `safeModuleVersion` to `0.3.0` when calling the `Safe4337Pack.init` method. This version of the Safe 4337 Module is the one compatible with the Entrypoint v0.7.

However, attempting to set `safeModuleVersion` to `0.3.0` in `SafeERC4337Pack` results in the following error:
Error: Incompatibility detected: Safe modules version 0.3.0 is not supported. The SDK can use 0.2.0 only.

### Fix
This PR resolves the issue by allowing `SafeERC4337Pack` to use version `0.3.0` as expected according to the documentation.

### DIFF
--- a/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
+++ b/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
@@ -51,7 +51,7 @@ import {
   signSafeOp,
   userOperationToHexValues
 } from './utils'
-import { entryPointToSafeModules, EQ_OR_GT_0_3_0 } from './utils/entrypoint'
+import { entryPointToSafeModules, GT_0_3_0 } from './utils/entrypoint'
 import { PimlicoFeeEstimator } from './estimators/PimlicoFeeEstimator'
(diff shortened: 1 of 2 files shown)


yagopv:
Hey @Another-DevX, this is already fixed in the [dev branch](https://github.com/safe-global/safe-core-sdk/blob/060a09491b81aa781ad71e1763200ca88c1f21bf/packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts#L137-L164)

Could you try with the latest [alpha package](https://www.npmjs.com/package/@safe-global/relay-kit/v/4.0.0-alpha.3
)

We have the 4.0.0 version pending to release

dasanra:
`relay-kit` `v4` was already published.

PR #1164, opened 2025-03-10
feat(abstract): Update config.ts

DenSmolonski:
## What it solves
Add Abstract short name


yagopv:
Hey, thanks! This should be added to the [@safe-global/safe-deployments](https://github.com/safe-global/safe-deployments) repository. I see it's already merged there, so we will update it here by bumping the dependencies. You can expect it to be included in our next release (Hopefully this week)

Issue #803, opened 2024-05-03
Multisend for calls only should not allow ˋDELEGATECALL` operation


clemsos:
## Description

When creating a multicall transaction with only calls, the operation should not return a ˋDELEGATECALL` operation, as the multisend function of the call only contract does not allow it.

When `onlyCalls` flag is set in `createTransaction`, the operation for the calls should always be 0.

## Environment

- Safe Core SDK version: 
- Safe contract version: 
- Environment:
  - non-browser: hardhat

## Steps to reproduce

- Parse a set of calls to a proxy using ˋcreateTransaction` with ˋonlyCalls` flag set
- Execute the tx in a safe

The tx will revert as the 1 operation is not allowed

## Expected result

Multisend should execute all txs properly.

## Additional context

Tested a with passing operation CALL to all packed txs and it worked, while using ˋoperation: 1ˋ fails.

PR #1172, opened 2025-03-14
feat(protocol-kit): set MultiSendCallOnly by default for MultiSend transactions

### LINKED ISSUE #1168: tx service return error - Operation DELEGATE_CALL is not allowed
## Description

using starter-kit send multi transactions got "Operation DELEGATE_CALL is not allowed"

## Environment
- Safe Core SDK versions:
  - sdk-starter-kit: 1.1.7
- Safe contract version: 
- Environment:
  - non-browser

## Steps to reproduce

when call `client.send` with multi txs(more than 1), the tx service return error "Operation DELEGATE_CALL is not allowed". And the generate safe transaction's `to` is Safe Official MultiSend contract: 0x38869bf66a61cF6bDB996A6aE40D5853Fd43B526

## Expected result

Please describe concisely what you expect instead.

## Additional context

how can i send multi txs using sdk?

dasanra:
## What it solves
Resolves #1168 and #803

## How this PR fixes it
Sets MultiSendCallOnly as default multisend operation

### DIFF
--- a/packages/protocol-kit/src/Safe.ts
+++ b/packages/protocol-kit/src/Safe.ts
@@ -68,6 +68,7 @@ import EthSafeTransaction from './utils/transactions/SafeTransaction'
 import { SafeTransactionOptionalProps } from './utils/transactions/types'
 import {
   encodeMultiSendData,
+  hasDelegateCalls,
   isNewOwnerPasskey,
   isOldOwnerPasskey,
   isPasskeyParam,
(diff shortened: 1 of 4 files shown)


yagopv:
Nice idea to throw the error if the batch contain delegate calls. Perhaps we should we add a test for the new error being throwed

yagopv (approved):

PR #1173, opened 2025-03-16
allow to specify isL1SafeSingleton and if to skipApproveTransaction

itsdeka:


itsdeka:
This is not the ideal solution. We should not override isL1SafeSingleton. It should exists instead an easier way to define the singleton gnosis safe address to be used for deployments.

Adding an option to skip the approve tx should be good instead.

PR #1188, opened 2025-04-01
feat: add info about Hoodi testnet to api-kit and protocol-kit

DenSmolonski:
This pull request adds necessary configuration for the Hoodi testnet to the api-kit and protocol-kit libraries, enabling proper interaction with the network.

Chainlist: https://chainlist.org/chain/560048
Safe transaction service: https://transaction-ethereum-hoodi.safe.protofire.io/


dasanra:
Added in #1199

PR #1191, opened 2025-04-03
refactor: simplify decodeSafeTxGas

PaulRBerg:
The current implementation is unnecessarily verbose.


dasanra:
Hey @PaulRBerg thank you for your PR. In this case we can't accept it as there are cases where this logic is needed. The error response depends on the chain/node provider and we had cases where the node was encoding the response appending 0x at the beggining and some that didn't. That's why we "split" the 0x to ensure none has it and then we add it again.

Then doing the `slice` to extract the exact data in one or two steps is a matter of choice, we just wanted to make explicit we are taking the 10 digits represeting the `safeTxGas` and that the index is in position 184 (when the string starts with 0x)

Maybe some comments would help to understand that piece of code better

PR #1209, opened 2025-05-05
feat: update safe module deployments version

Another-DevX:
This PR updates the `@safe-global/relay-kit dependency` to the latest version in order to support the EntryPoint v0.7 on the CELO network.


dasanra:
Close for #1210

Issue #1214, opened 2025-05-16
Add api keys support to Api Kit


yagopv:
## Context / issue
We are adding Api keys as a mandatory requisite to use the Transaction Service

## Proposed solution
- [ ] Adding Api keys support to the api-kit (Authorization header)
- [ ] Update transaction service endpoints

PR #1216, opened 2025-05-19
feat(api-kit): Api keys integration

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-core-sdk/issues/1214

## How this PR fixes it
- Update production Transaction Service URL
- Add `Authorization` header to the `api-kit` requests
- Update tests using the `SafeApiKit` class to include the api key field
- Update github action for `api-kit` tests to create the required .env file


packages/api-kit/tests/.env-sample
```diff
@@ -0,0 +1,2 @@
+# Transaction Service API Key
+TX_SERVICE_API_KEY=
```
dasanra, line 2: This file should be at `api-kit` top level, not inside `tests`folder

yagopv: done

PR #1218, opened 2025-05-24
4337 GenericFeeEstimator than can work with multiple providers

sherifahmed990:
## What it solves
Resolves #
Allow relay-kit to work with multiple 4337 bundler and paymaster providers

## How this PR fixes it
by implementing GenericFeeEstimator


sherifahmed990:

packages/relay-kit/src/packs/safe-4337/estimators/generic/types.ts
```diff
@@ -0,0 +1,36 @@
...
+          paymaster?: string
+          paymasterData?: string
+          paymasterVerificationGasLimit?: string
+          paymasterPostOpGasLimit?: string
+          isFinal?: boolean
+        }
+  }
+]
+
+export type PaymasterRpcSchema = [
```
yagopv, line 23: Do you really need this being part of the standard?

yagopv: If you remove the generic from createBundlerClient it should work anyways

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -87,6 +87,7 @@ export class Safe4337Pack extends RelayKitBasePack<{
 
   #onchainIdentifier: string = ''
 
+  #nodeUrl?: string
```
yagopv, line 90: This rpcUrl can be obtained in several ways instead adding an additional parameter. It can be the `provider`in the initialization method or it can be extracted using the `SafeProvider` in the `protocol-kit` by retrieving the viem instance:

`protocolKit.getSafeProvider().getExternalProvider().chain?.rpcUrls[0]`

packages/relay-kit/src/packs/safe-4337/types.ts
```diff
@@ -144,6 +146,8 @@ export type EstimateFeeFunctionProps = {
   bundlerUrl: string
   entryPoint: string
   paymasterOptions?: PaymasterOptions
+  nodeUrl?: string
+  chainId?: bigint
```
yagopv, line 150: You can use `eth_chainId` instead passing it around

packages/relay-kit/src/packs/safe-4337/estimators/generic/types.ts
sherifahmed990: I will remove it.

packages/relay-kit/src/packs/safe-4337/estimators/generic/GenericFeeEstimator.ts
```diff
@@ -0,0 +1,179 @@
...
+  overrides: GenericFeeEstimatorOverrides
+
+  constructor(overrides: GenericFeeEstimatorOverrides = {}) {
+    this.defaultVerificationGasLimitOverhead =
+      overrides.defaultVerificationGasLimitOverhead ?? 35_000n
+    this.overrides = overrides
+  }
+
+  async preEstimateUserOperationGas({
+    bundlerUrl, // eslint-disable-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
sherifahmed990: there is no standard to retrieve the gas fees from the bundler, so we need a separate parameter for the nodeUrl

packages/relay-kit/src/packs/safe-4337/types.ts
```diff
@@ -29,6 +29,7 @@ type PredictedSafeOptions = {
 export type SponsoredPaymasterOption = {
   isSponsored: true
   sponsorshipPolicyId?: string
+  paymasterContext?: Record<string, any>
```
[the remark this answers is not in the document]
sherifahmed990: Yes, i will update it

packages/relay-kit/src/packs/safe-4337/types.ts
sherifahmed990: the chain id is a fixed value, calling eth_chainId will add unnecessary latency.

packages/relay-kit/src/packs/safe-4337/estimators/generic/GenericFeeEstimator.ts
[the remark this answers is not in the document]
yagopv: Yes, but I mean the `relay-kit` assumes you are working with bundlers and a paymasters, so URLs should be available in the `Safe4337Pack` initialization. In this case, the specific `GenericFeeEstimator` is not working with a bundler URL but is using an RPC URL instead for its needs. What do you think about adding the `rpcUrl` as a parameter in the `GenericFeeEstimator` constructor instead adding it to the interface functions for this only use case ?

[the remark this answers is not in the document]
sherifahmed990: I think passing the rpcUrl as a constructor param was my initial implementation.
But if the GenericFeeEstimator constructor has required parameters, it can't be set as the default estimator if you decided to do so in the future.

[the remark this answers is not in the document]
yagopv: it will be not the default at least medium term. But if we do that then we will add the rpc for sure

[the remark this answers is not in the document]
sherifahmed990: done

packages/relay-kit/src/packs/safe-4337/types.ts
```diff
@@ -156,6 +158,7 @@ export type EstimateFeeFunction = ({
 export interface IFeeEstimator {
   preEstimateUserOperationGas?: EstimateFeeFunction
   postEstimateUserOperationGas?: EstimateFeeFunction
+  defaultVerificationGasLimitOverhead?: bigint
```
yagopv, line 161: This could be as well a constructor parameter for the estimator right? Instead adding it to the main `Safe4337Pack` estimation method

packages/relay-kit/src/packs/safe-4337/Safe4337Pack.ts
```diff
@@ -409,11 +409,17 @@ export class Safe4337Pack extends RelayKitBasePack<{
     feeEstimator = new PimlicoFeeEstimator()
   }: EstimateFeeProps): Promise<BaseSafeOperation> {
     const threshold = await this.protocolKit.getThreshold()
+    const nodeUrl = this.protocolKit.getSafeProvider().getExternalProvider().transport.url
```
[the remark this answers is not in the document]
sherifahmed990: done

packages/relay-kit/src/packs/safe-4337/types.ts
sherifahmed990: I can add the defaultVerificationGasLimitOverhead logic to postEstimateUserOperationGas, but
i would need to add threshold to EstimateFeeFunctionProps.

sherifahmed990: done

yagopv:
So thanks @sherifahmed990 for your contribution :) 

Will continue [here](https://github.com/safe-global/safe-core-sdk/pull/1238)

PR #1220, opened 2025-06-04
[FIX] pass through onchainAnalytics

nlparis:
## What it solves
Adds onchain tracking when you connect a signer using the .connect() method. Otherwise, `onchainAnalytics` are not available when executing.

https://docs.safe.global/sdk/onchain-tracking

## How this PR fixes it
Allows signer to use the onchain tracking after connecting a signer to an initialized safe before execution.


nlparis:

nlparis:
recheck

PR #1245, opened 2025-07-16
chore: bump safe deployments

yagopv:
## What it solves

- Bump Safe Deployments to the latest version
- Update scripts to use `https://chainlist.org/rpcs.json` as the default list with fallback to `ethereum-lists/chains`


yagopv:
Should we update shortnames here based on he new source of truth ?

dasanra (approved):

PR #1333, opened 2026-04-28
fix: scope abitype AddressType overrides to SDK builds

### LINKED ISSUE #1323: @safe-global/types-kit globally pollutes abitype AddressType to `string`
## Problem

`@safe-global/types-kit` exports a global `declare module 'abitype'` that overrides `AddressType` to `string`:

https://github.com/safe-global/safe-core-sdk/blob/main/packages/types-kit/src/index.ts#L15-L24

```ts
declare module 'abitype' {
  export interface Register {
    AddressType: string
  }
}
```

This is a **global ambient declaration** — any project that depends on `@safe-global/types-kit` (even transitively via `api-kit` or `protocol-kit`) has its abitype `Address` type silently widened from `` `0x${string}` `` to `string`.

## Impact

- **All viem/wagmi types are affected.** `useAccount().address` becomes `string | undefined` instead of `` `0x${string}` | undefined ``, breaking type-safety for any project that uses both wagmi and Safe SDK.
- The consumer has **no way to opt out** short of adding their own counter-declaration in every TypeScript compilation unit:
  ```ts
  declare module 'abitype' {
    interface Register {
      AddressType: `0x${string}`;
    }
  }
  ```
- This affects any monorepo or app that pulls in `@safe-global/api-kit` or `@safe-global/protocol-kit` as a dependency, even if the Safe SDK is only used in a small part of the codebase.

## Suggestion

Move the `declare module 'abitype'` block out of the published package's type declarations. If Safe SDK needs `AddressType: string` internally (e.g. for ethers.js compatibility), it should be scoped to the SDK's own compilation — not exported to consumers.

Options:
1. Move it to a `tsconfig` or internal `.d.ts` file that is **not** included in the published `dist/`
2. Use type assertions (`as string`) at the Safe SDK boundary instead of globally overriding abitype's Register

## Environment

- `@safe-global/types-kit`: 3.0.0
- `abitype`: 1.2.3
- `viem`: 2.44.4
- `wagmi`: 2.19.5
- `typescript`: 5.9.3

Genmin:
## Summary
- Move the Safe SDK `abitype` `AddressType: string` augmentation out of public package entrypoints
- Keep the override in package-local non-emitted `src/abitype.d.ts` files so internal builds continue to accept existing string address usage
- Cover `types-kit`, `protocol-kit`, `api-kit`, `relay-kit`, and `sdk-starter-kit` so consumers are no longer affected transitively

## Why
The published `dist/src/index.d.ts` entrypoints currently include `declare module 'abitype'`, which widens `viem`/`abitype` `Address` from `` `0x${string}` `` to `string` for any downstream app that imports Safe packages. This keeps the SDK-local compatibility shim but stops exporting it into consumer TypeScript programs.

Closes #1323.

## Validation
- `yarn build`
- targeted ESLint on changed files
- targeted Prettier check on changed files
- scanned generated `packages/*/dist` declarations for `declare module 'abitype'` / `AddressType: string`
- external consumer `tsc --noEmit` fixture importing built Safe package declarations and verifying `viem` `Address` remains narrow

### DIFF
--- a/packages/api-kit/src/abitype.d.ts
+++ b/packages/api-kit/src/abitype.d.ts
@@ -1,0 +1,7 @@
+import 'abitype'
+
+declare module 'abitype' {
+  export interface Register {
+    AddressType: string
+  }
+}
--- a/packages/api-kit/src/index.ts
+++ b/packages/api-kit/src/index.ts
@@ -3,9 +3,3 @@ import SafeApiKit, { SafeApiKitConfig } from './SafeApiKit'
 export * from './types/safeTransactionServiceTypes'
 export { SafeApiKitConfig }
 export default SafeApiKit
-
-declare module 'abitype' {
-  export interface Register {
-    AddressType: string
-  }
-}
--- a/packages/protocol-kit/src/abitype.d.ts
+++ b/packages/protocol-kit/src/abitype.d.ts
@@ -1,0 +1,7 @@
+import 'abitype'
+
+declare module 'abitype' {
+  export interface Register {
+    AddressType: string
+  }
+}
--- a/packages/protocol-kit/src/index.ts
+++ b/packages/protocol-kit/src/index.ts
@@ -151,9 +151,3 @@ export {
 export * from './types'
 
 export default Safe
-
-declare module 'abitype' {
-  export interface Register {
-    AddressType: string
-  }
-}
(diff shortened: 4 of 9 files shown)


dasanra:
Fixed by #1324

PR #1336, opened 2026-05-07
fix(protocol-kit): refine baseGas estimation against execTransaction

Uxio0:
## Summary

Aligns `estimateTxBaseGas` with the current Safe `execTransaction` implementation so refunds (especially ERC20) and event emission are not under-counted, which could lead to relayer-side reverts.

- Replace the flat `TRANSFER_GAS_COST = 32_000` with `calculateRefundGas`, which models `handlePayment` per gas token:
  - **native ETH**: cold address access (EIP-2929) + value transfer, plus NEWACCOUNT (EIP-161) when the receiver is a fresh account
  - **ERC20**: typical transfer cost, plus the SSTORE 0→non-zero delta (EIP-2200) when the receiver has no prior token balance (detected via `balanceOf`; reverts treated as new-holder, conservative)
  - returns `0` when `gasPrice == 0` (`handlePayment` is gated by it)
- Add `calculateExecTransactionEventsGas` accounting for `ExecutionSuccess`/`ExecutionFailure` (always emitted) and `SafeMultiSigTransaction` (L2 singleton only).
- Drop the stale pre-Istanbul `+64/+128` adjustment for the `baseGas` calldata placeholder, replacing it with the correct delta based on the byte count of the **final** value, applied at the end of the function so the count reflects all add-ons.
- Break the previously-bundled overhead into `INTRINSIC_TX_GAS_COST` + `PRE_EXEC_STORAGE_GAS_COST` + `MISC_OVERHEAD_GAS_COST` for traceability and headroom.
- Add a TODO for transaction guard hooks (cost depends on the guard implementation).

## Net change vs. pre-refactor

| Scenario | Old `baseGas` add-ons | New `baseGas` add-ons | Δ |
|---|---|---|---|
| ETH refund, existing receiver | 32 000 | 28 200 + 11 600 + 1 500 = **41 300** | +9 300 |
| ETH refund, fresh receiver | 32 000 + 25 000 = 57 000 | 28 200 + 36 600 + 1 500 = **66 300** | +9 300 |
| ERC20, existing token holder | 32 000 | 28 200 + 21 000 + 1 500 = **50 700** | +18 700 |
| ERC20, new token holder | 32 000 | 28 200 + 38 000 + 1 500 = **67 700** | +35 700 |
| `gasPrice = 0` | 32 000 (+ 25k if new ETH) | 28 200 + 0 + 1 500 = **29 700** | -2 300 / -27 300 |

The `gasPrice = 0` row is the only place we go *below* the old number. That is correct: the old code charged `TRANSFER_GAS_COST` (and even `NEW_ACCOUNT_GAS_COST`) when `handlePayment` was never invoked.

Refs: [PLA-1429](https://linear.app/safe-global/issue/PLA-1429/fix-base-gas-estimation-for-gtf)

### DIFF
--- a/packages/protocol-kit/src/utils/transactions/gas.ts
+++ b/packages/protocol-kit/src/utils/transactions/gas.ts
@@ -177,6 +177,9 @@ async function hasExistingTokenBalance(
  *   (EIP-161) when the receiver is a fresh account.
  * - ERC20 refund: typical transfer cost; adds extra gas (SSTORE 0->non-zero per EIP-2200)
  *   when the receiver has no prior balance for the token.
+ *
+ * Note: underestimations are expected if `gasToken` is set and ERC20
+ * contract is a proxy or includes additional logic
  */
 async function calculateRefundGas(
   safeProvider: SafeProvider,
--- a/packages/protocol-kit/src/utils/transactions/gas.ts
+++ b/packages/protocol-kit/src/utils/transactions/gas.ts
@@ -183,7 +183,7 @@ async function hasExistingTokenBalance(
  */
 async function calculateRefundGas(
   safeProvider: SafeProvider,
-  gasPrice: number | string,
+  gasPrice: string,
   gasToken: string,
   refundReceiver: string
 ): Promise<number> {
--- a/packages/protocol-kit/src/utils/transactions/gas.ts
+++ b/packages/protocol-kit/src/utils/transactions/gas.ts
@@ -124,8 +124,7 @@ function calculateExecTransactionEventsGas(
     const additionalInfoDynamicBytes = 32 + 3 * 32 // (nonce, msg.sender, threshold)
     const eventDataBytes =
       headBytes + dataDynamicBytes + signaturesDynamicBytes + additionalInfoDynamicBytes
-    gas +=
-      LOG_BASE_GAS_COST + LOG_TOPIC_GAS_COST + eventDataBytes * LOG_DATA_GAS_COST_PER_BYTE + 500 // additionalInfo abi.encode + memory expansion
+    gas += LOG_BASE_GAS_COST + LOG_TOPIC_GAS_COST + eventDataBytes * LOG_DATA_GAS_COST_PER_BYTE
   }
 
   return gas
@@ -163,7 +162,7 @@ async function hasExistingTokenBalance(
     args: [account as `0x${string}`]
   })
   try {
-    const response = await safeProvider.call({ to: gasToken, value: '0', data })
+    const response = await safeProvider.call({ to: gasToken, from: gasToken, value: '0', data })
     return BigInt(response || '0x0') > 0n
   } catch {
     // Reverts and empty responses are treated as "no balance" (conservative: higher cost)
--- a/packages/protocol-kit/src/utils/transactions/gas.ts
+++ b/packages/protocol-kit/src/utils/transactions/gas.ts
@@ -1,4 +1,10 @@
-import { BaseError, CallExecutionErrorType, RawContractErrorType } from 'viem'
+import {
+  BaseError,
+  CallExecutionErrorType,
+  RawContractErrorType,
+  encodeFunctionData,
+  parseAbi
+} from 'viem'
 import { OperationType, SafeVersion, SafeTransaction } from '@safe-global/types-kit'
 import semverSatisfies from 'semver/functions/satisfies.js'
 import Safe from '@safe-global/protocol-kit/Safe'
@@ -24,8 +30,8 @@ const CALL_DATA_ZERO_BYTE_GAS_COST = 4
 // Every byte != 00 -> 16 Gas cost (68 before Istanbul)
 const CALL_DATA_BYTE_GAS_COST = 16
 
-// gas cost initialization of a Safe
-const INITIZATION_GAS_COST = 20_000
+// gas cost initialization of a Safe (SSTORE 0→non-zero on a cold slot = 20 000 + 2 100 cold-load surcharge per EIP-2929)
+const INITIALIZATION_GAS_COST = 22_100
 
 // increment nonce gas cost
 const INCREMENT_NONCE_GAS_COST = 5_000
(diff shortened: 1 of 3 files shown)


packages/protocol-kit/src/utils/transactions/gas.ts
```diff
@@ -24,8 +30,8 @@ const CALL_DATA_ZERO_BYTE_GAS_COST = 4
 // Every byte != 00 -> 16 Gas cost (68 before Istanbul)
 const CALL_DATA_BYTE_GAS_COST = 16
 
-// gas cost initialization of a Safe
-const INITIZATION_GAS_COST = 20_000
+// gas cost initialization of a Safe (SSTORE 0→non-zero on a cold slot = 20 000 + 2 100 cold-load surcharge per EIP-2929)
+const INITIALIZATION_GAS_COST = 22_100
```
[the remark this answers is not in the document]
Uxio0: This is only relevant for <= 1.3.0, not relevant

packages/protocol-kit/src/utils/transactions/gas.ts
```diff
@@ -84,6 +152,55 @@ async function isNewEthRefundReceiver(
...
+ *   when the receiver has no prior balance for the token.
+ */
+async function calculateRefundGas(
+  safeProvider: SafeProvider,
+  gasPrice: number | string,
+  gasToken: string,
+  refundReceiver: string
+): Promise<number> {
+  if (BigInt(gasPrice) === 0n) {
+    return 0
```
dasanra, line 191: If refundReceiver is set but gasPrice isn't user may expect that gas is correctly calculated, but this silently returns 0 which can lead to wrong estimations as neither gasPrice or refundReceiver are mandatory parameters.

packages/protocol-kit/src/utils/transactions/gas.ts
```diff
@@ -84,6 +155,55 @@ async function isNewEthRefundReceiver(
...
+  }
+
+  if (gasToken.toLowerCase() === ZERO_ADDRESS.toLowerCase()) {
+    let gas = COLD_ACCOUNT_ACCESS_GAS_COST + CALL_VALUE_GAS_COST
```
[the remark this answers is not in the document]
Uxio0: Not very relevant, tx.origin does not have to match tx.sender and overpayment in that case is minimal

packages/protocol-kit/src/utils/transactions/gas.ts
```diff
@@ -84,6 +151,55 @@ async function isNewEthRefundReceiver(
...
+    if (await isNewEthRefundReceiver(safeProvider, gasToken, refundReceiver)) {
+      gas += NEW_ACCOUNT_GAS_COST
+    }
+    return gas
+  }
+
+  const hasBalance = await hasExistingTokenBalance(safeProvider, gasToken, refundReceiver)
+  return hasBalance ? ERC20_TRANSFER_GAS_COST : ERC20_NEW_HOLDER_TRANSFER_GAS_COST
```
akshay-ap, line 203: I think this still does not cover cases:
- when token is a proxy contract
    - `SLOAD ` - implementation address
    - `delegatecall` - to implementation
- any additional token specific logic. e.g. USDC has 2 blacklist checks in each transfer
- PYUSD is also a proxy and checks if sender is admin (another `SLOAD`), and then loads the implementation contract + check for receiver/sender are frozen (2 `SLOAD`)

akshay-ap: How about executing an rpc call to estimate actual gas costs of the transfer rather than calculating based on instructions executed?
If rpc call fails fallback to the manual estimations.
For different tokens there will be different estimates. For example below are some of the tokens on mainnet and their gas estimates when sending various Erc20 tokens from 1/1 Safe to an EOA.


Chain Id: 1
Sender: `0xff34811dEA6154E41Fe6b0FEF4d8B60C5A334854`
Recipient: `0x8b0aB586dF1Ca1f360cb26b34eEC2C3AF969E821` (initial balance of this address for all tokens > 0)
Amount: 1 wei

| Symbol | Address | Decimals | estimateGas | approx transferOnly (− 21k intrinsic) | Notes |
|---|---|---|---:|---:|---|
| DAI | `0x6B175474E89094C44Da98b954EedeAC495271d0F` | 18 | 35005 | 14005 | — |
| USDC | `0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48` | 6 | 45504 | 24504 | — |
| WETH | `0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2` | 18 | 34793 | 13793 | — |
| PYUSD | `0x6c3ea9036406852006290770BEdFcAbA0e23A0e8` | 6 | 48939 | 27939 | — |
| TEST | `0x58e48F88bF943367f05F31164d2943FF4008EEc6` | 18 | 35360 | 14360 | — |


PYUSD tenderly simulation: https://www.tdly.co/shared/simulation/66352e06-cf84-4573-9136-474c877d5898

packages/protocol-kit/src/utils/transactions/gas.ts
Uxio0: `handlePayment` is only called if `gasPrice > 0`: https://github.com/safe-fndn/safe-smart-account/blob/main/contracts/Safe.sol#L219

packages/protocol-kit/src/utils/transactions/gas.ts
Uxio0: It's a really good idea, but I feel it would be very difficult to debug. I would start like this and then analyze if our estimations are close to what's expected or not and revisit this

akshay-ap: I think we should at least document that underestimations are expected when the contract is a proxy or includes additional logic beyond the bare minimum required to implement the ERC20 spec.

Uxio0: Agree, done in 3626eeb8

packages/protocol-kit/src/utils/transactions/gas.ts
```diff
@@ -84,6 +151,58 @@ async function isNewEthRefundReceiver(
...
+ *   (EIP-161) when the receiver is a fresh account.
+ * - ERC20 refund: typical transfer cost; adds extra gas (SSTORE 0->non-zero per EIP-2200)
+ *   when the receiver has no prior balance for the token.
+ *
+ * Note: underestimations are expected if `gasToken` is set and ERC20
+ * contract is a proxy or includes additional logic
+ */
+async function calculateRefundGas(
+  safeProvider: SafeProvider,
+  gasPrice: number | string,
```
dasanra: gasPrice should be only string

```suggestion
  gasPrice: string,
```

Uxio0:
Agree @dasanra , done

dasanra (approved):

PR #1339, opened 2026-05-17
support packed P256.Verifiers format in isSharedSigner

ivanmmurciaua:
## What it solves
Resolves # (no prior issue — discovered during EIP-7951 precompile integration)

`isSharedSigner` returns `false` for any Safe whose `verifiers` field uses the packed `P256.Verifiers` encoding `(uint176(precompile) << 160) | uint176(uint160(fallback))`, which is the canonical format documented by Safe to activate the EIP-7951 P-256 precompile. Strict equality between a 176-bit packed value and a 160-bit address always fails, causing `createPasskeyClient` to fall back to the factory signer path and `signSafeOperation` to throw **"UserOperations can only be signed by Safe owners"** with no indication that the verifier encoding is the cause.

## How this PR fixes it

Masks the stored `verifiers` to the lower 160 bits before comparing against the expected fallback address. This handles both encodings with a single branch: when the upper bits are zero (simple format) the mask is a no-op; when they encode a precompile slot (packed format) the mask extracts the fallback verifier.

Also adds a defensive `BigInt` cast since viem returns `bigint` but ethers v5 and test mocks may return `string` or `BigNumber`. 10 unit tests added covering both encoding formats, coordinate mismatches, and the missing-SharedSigner-in-owners guard.

### DIFF
--- a/packages/protocol-kit/src/utils/passkeys/isSharedSigner.ts
+++ b/packages/protocol-kit/src/utils/passkeys/isSharedSigner.ts
@@ -29,10 +29,20 @@ async function isSharedSigner(
     const verifierAddress =
       passkey.customVerifierAddress || getDefaultFCLP256VerifierAddress(chainId)
 
+    // P256.Verifiers (uint176) packs two values: upper 16 bits = precompile slot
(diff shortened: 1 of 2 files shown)


ivanmmurciaua:

PR #1348, opened 2026-05-30
Fix numeric block id handling

samsamtrum:
Fixes numeric `blockTag` handling in `SafeProvider` helpers.

`SafeProvider.getBalance`, `getNonce`, `getContractCode`, `isContractDeployed`, and `call` accept `blockTag?: string | number`, but numeric values were passed through `asBlockId()` to `blockNumber()`, which called `.toNumber()` on a JavaScript number. That throws before the viem action is called.

This returns the numeric block id directly and adds a small unit test for numeric and string block ids.

Tested:

```bash
cd packages/protocol-kit
../../node_modules/.bin/mocha -r ts-node/register -r tsconfig-paths/register tests/unit/block.test.ts
```


dasanra:
Thank you for your contribution, fixing it more widely in #1366

PR #1378, opened 2026-06-17
chore: upgrade viem to 2.52.2 and keep chainId as bigint

Uxio0:
Bump viem from 2.21.8 to 2.52.2 across all packages.

- chainId stays a bigint through the EIP-712 hashing helpers and the signing path instead of being cast to number, which lost precision for chain IDs above Number.MAX_SAFE_INTEGER. viem 2.23+ accepts a bigint chainId in the domain, so no shim is needed.
- viem 2.52 tightened ABI types to 0x${string}. The SDK keeps its public API on plain string (abitype AddressType override) and casts to 0x${string} only at the viem call boundaries, so consumer types are unchanged.
- protocol-kit's declaration build uses moduleResolution bundler (viem's documented requirement) so viem/ox resolve to their shipped .d.ts. This is build-internal: emitted declarations are byte-identical and consumers are unaffected.

Closes PLA-1668


dasanra (approved):
Did some changes about unnecessary casts aligning viem and abitype dependencies. Now it looks good.

yagopv (approved):

PR #1418, opened 2026-08-31
fix(protocol-kit): add isSignedBy for checksum-safe signature lookup

### LINKED ISSUE #495: addSignature doesn't checksum addresses
Keys in the signatures Map are lower-case, non-checksummed, addresses.

gomesalexandre:
closes #495

tl;dr `addSignature()` stores signer addresses lowercased in the `signatures` Map internally, but that Map is exposed publicly - so `signatures.has(checksummedAddress)` silently returns `false` unless the caller happens to pass an already-lowercased address. `getSignature(signer)` already normalizes its own input (`signer.toLowerCase()`) before comparing, so it was never actually broken for checksummed lookups - only direct Map access was.

## the fix

Adds `isSignedBy(signer: string): boolean` to `EthSafeTransaction` and `EthSafeMessage` (and their shared `SafeTransaction`/`SafeMessage` interfaces in `types-kit`), matching the API shape suggested in the issue itself. Internally it's just `this.signatures.has(signer.toLowerCase())` - correct because lowercasing an EIP-55 checksummed address never changes which underlying address it represents (checksum casing is a display/validation layer on top of the same 20 bytes), so there's no risk of two different addresses colliding via lowercase comparison.

Didn't touch the raw `signatures` Map's key casing itself (kept lowercase) rather than changing it, to avoid any risk to existing code that might already rely on the current lowercase-key behavior when iterating the Map directly.

## one thing worth a maintainer call

`isSignedBy` was added to the public `SafeTransaction`/`SafeMessage` interfaces in `types-kit`, not just the concrete classes. Within this repo only `EthSafeTransaction`/`EthSafeMessage` implement them, so nothing here breaks - but if there's any external code implementing these interfaces directly (rather than extending these classes), this would be a new required method for them too. Flagging in case that matters for your semver conventions; happy to adjust the approach if so.

## receipts

New unit tests (checksummed address, already-lowercased address, all-uppercase address, and a non-signer address all correctly recognized/rejected):

```
$ npx mocha -r ts-node/register -r tsconfig-paths/register tests/unit/SafeTransaction.test.ts tests/unit/SafeMessage.test.ts

  EthSafeTransaction
    addSignature/isSignedBy checksum handling
      ✔ recognizes a signer via isSignedBy regardless of address casing
      ✔ isSignedBy returns false for an address that never signed
      ✔ the raw signatures Map still uses lowercase keys internally

  EthSafeMessage
    addSignature/isSignedBy checksum handling
      ✔ recognizes a signer via isSignedBy regardless of address casing
      ✔ isSignedBy returns false for an address that never signed
      ✔ the raw signatures Map still uses lowercase keys internally

  6 passing (3ms)
```

Full `protocol-kit` unit suite, no regressions:

```
$ npx mocha -r ts-node/register -r tsconfig-paths/register 'tests/unit/**/*.ts'
  46 passing (22ms)
```

`protocol-kit` and `types-kit` typecheck (`build:types`) clean, `eslint` clean on all changed files.

Risk: low - purely additive method, no existing behavior changed.

### DIFF
--- a/.changeset/isSignedBy-checksum-fix.md
+++ b/.changeset/isSignedBy-checksum-fix.md
@@ -1,0 +1,6 @@
+---
+"@safe-global/protocol-kit": patch
+"@safe-global/types-kit": patch
+---
+
+Add `isSignedBy(signer)` to `SafeTransaction`/`SafeMessage` for checksum-safe signature lookups. The `signatures` Map stores signer addresses lowercased internally, so `signatures.has(checksummedAddress)` always returned `false` unless the caller passed an already-lowercased address - `isSignedBy` normalizes the input before checking.
--- a/packages/protocol-kit/src/utils/messages/SafeMessage.ts
+++ b/packages/protocol-kit/src/utils/messages/SafeMessage.ts
@@ -17,6 +17,10 @@ class EthSafeMessage implements SafeMessage {
     this.signatures.set(signature.signer.toLowerCase(), signature)
   }
 
+  isSignedBy(signer: string): boolean {
+    return this.signatures.has(signer.toLowerCase())
+  }
+
   encodedSignatures(): string {
     return buildSignatureBytes(Array.from(this.signatures.values()))
   }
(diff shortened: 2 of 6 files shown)


gomesalexandre:

gomesalexandre:
recheck

PR #1426, opened 2026-09-07
fix: keep TransactionOptions.nonce 0 when converting execution options

### LINKED ISSUE #1425: TransactionOptions nonce: 0 is dropped when converting execution options
## Summary

`TransactionOptions.nonce = 0` is currently dropped when Protocol Kit converts execution options for both legacy and EIP-1559 transactions. This makes the SDK ignore an explicit caller-supplied nonce of zero, even though `0` is a valid EOA transaction nonce.

This looks like a small falsy-value handling issue rather than a contract-level security issue, but it can surprise callers in local/dev chains, freshly funded signer accounts, deterministic transaction tests, and any flow that intentionally executes the first transaction from an owner with `nonce: 0`.

## Affected code

In `packages/protocol-kit/src/utils/transactions/utils.ts`, both conversion helpers guard the nonce with a truthiness check:

```ts
export function createLegacyTxOptions(
  options?: TransactionOptions
): Partial<WalletLegacyTransactionOptions> {
  const converted: Partial<WalletLegacyTransactionOptions> = {}
  if (options?.from) {
    converted.account = options.from
  }

  if (options?.gasLimit) {
    converted.gas = BigInt(options.gasLimit)
  }

  if (options?.gasPrice) {
    converted.gasPrice = BigInt(options.gasPrice)
  }

  if (options?.nonce) {
    converted.nonce = options.nonce
  }

  return converted
}

export function createTxOptions(options?: TransactionOptions): Partial<WalletTransactionOptions> {
  const converted: Partial<WalletTransactionOptions> = {}
  if (options?.from) {
    converted.account = options.from
  }

  if (options?.gasLimit) {
    converted.gas = BigInt(options.gasLimit)
  }

  if (options?.maxFeePerGas) {
    converted.maxFeePerGas = BigInt(options.maxFeePerGas)
  }

  if (options?.maxPriorityFeePerGas) {
    converted.maxPriorityFeePerGas = BigInt(options.maxPriorityFeePerGas)
  }

  if (options?.nonce) {
    converted.nonce = options.nonce
  }

  return converted
}
```

Because `0` is falsy in JavaScript/TypeScript, `convertTransactionOptions({ nonce: 0, ... })` returns an object without `nonce`.

## Why this matters

The public execution API accepts `TransactionOptions`:

```ts
async executeTransaction(
  safeTransaction: SafeTransaction | SafeMultisigTransactionResponse,
  options?: TransactionOptions
): Promise<TransactionResult> {
  ...
  const txResponse = await this.#contractManager.safeContract.execTransaction(
    signedSafeTransaction,
    {
      from: signerAddress,
      ...options
    }
  )
  return txResponse
}
```

The conversion layer later maps those options into viem wallet transaction parameters. For `nonce: 123456789`, the current e2e test already verifies that the option is honored:

```ts
const execOptions: TransactionOptions = { nonce: 123456789 }
await chai
  .expect(safeSdk1.executeTransaction(tx, execOptions))
  .to.be.rejectedWith('Nonce too high')
```

However, the edge case `nonce: 0` is not covered. If a caller explicitly passes `nonce: 0`, the SDK silently omits it and lets the wallet/provider choose the nonce instead.

## Suggested fix

Use an existence check instead of a truthiness check in both helpers:

```ts
if (typeof options?.nonce !== 'undefined') {
  converted.nonce = options.nonce
}
```

or:

```ts
if (options && 'nonce' in options) {
  converted.nonce = options.nonce
}
```

It would also be useful to add a regression test for:

```ts
expect(createTxOptions({ nonce: 0, ...requiredFields }).nonce).toBe(0)
expect(createLegacyTxOptions({ nonce: 0, ...requiredFields }).nonce).toBe(0)
```

I noticed this while reviewing account-abstraction/wallet SDK option handling paths. Happy to provide a small PR if that would be useful.

devtechedge:
## What it solves
Resolves #1425

`TransactionOptions.nonce = 0` is a valid EOA nonce (fresh signer, local/dev chain, first tx). Protocol Kit was dropping it because the conversion helpers treated `0` as falsy, so `executeTransaction(..., { nonce: 0 })` never reached viem and the wallet/provider picked a nonce instead.

## How this PR fixes it

Public execution still spreads caller options into `execTransaction`:

```ts
async executeTransaction(
  safeTransaction: SafeTransaction | SafeMultisigTransactionResponse,
  options?: TransactionOptions
): Promise<TransactionResult>
```

`BaseContract` maps those options with `convertTransactionOptions(options)` in `packages/protocol-kit/src/utils/transactions/utils.ts`. That helper dispatches on `isLegacyTransaction` (`!!options?.gasPrice`) to either `createLegacyTxOptions` or `createTxOptions`. Both used:

```ts
if (options?.nonce) {
  converted.nonce = options.nonce
}
```

Non-zero values worked — the existing e2e already checks `{ nonce: 123456789 }` (`Nonce too high`). `{ nonce: 0 }` was omitted. The same helpers feed `toTransactionRequest` and `toEstimateGasParameters`, so gas estimation and the wallet tx request had the same hole.

Fix: copy nonce with an existence check in both helpers.

```ts
if (typeof options?.nonce !== 'undefined') {
  converted.nonce = options.nonce
}
```

- `{ nonce: 0 }` is kept
- `{ nonce: 123 }` is unchanged
- omitted `nonce` stays omitted (wallet/provider still chooses)

Left alone on purpose:
- `gasLimit` / `gasPrice` / `maxFeePerGas` / `maxPriorityFeePerGas` still use truthiness. `0` is not a useful gas value.
- `isLegacyTransaction` still keys off `gasPrice`, so `{ nonce: 0 }` without `gasPrice` stays on the EIP-1559 path.
- The existing non-zero nonce e2e is not touched.

Unit tests in `packages/protocol-kit/tests/unit/tx-options.test.ts`:
- `createTxOptions({ nonce: 0 }).nonce === 0`
- `createTxOptions({ nonce: 123 }).nonce === 123`
- omitted nonce is `undefined` (`{}` and no-arg)
- same for `createLegacyTxOptions` (with `gasPrice` so the legacy helper is the one under test)
- `convertTransactionOptions({ nonce: 0 })` on both EIP-1559 and legacy paths

Patch changeset on `@safe-global/protocol-kit`.

### DIFF
--- a/.changeset/tx-options-nonce-zero.md
+++ b/.changeset/tx-options-nonce-zero.md
@@ -1,0 +1,5 @@
+---
+'@safe-global/protocol-kit': patch
+---
+
+Preserve `TransactionOptions.nonce = 0` when converting execution options so a caller-supplied zero nonce is not dropped.
--- a/packages/protocol-kit/src/utils/transactions/utils.ts
+++ b/packages/protocol-kit/src/utils/transactions/utils.ts
@@ -285,7 +285,7 @@ export function createLegacyTxOptions(
     converted.gasPrice = BigInt(options.gasPrice)
   }
 
-  if (options?.nonce) {
+  if (typeof options?.nonce !== 'undefined') {
     converted.nonce = options.nonce
   }
 
@@ -310,7 +310,7 @@ export function createTxOptions(options?: TransactionOptions): Partial<WalletTra
     converted.maxPriorityFeePerGas = BigInt(options.maxPriorityFeePerGas)
   }
 
-  if (options?.nonce) {
+  if (typeof options?.nonce !== 'undefined') {
     converted.nonce = options.nonce
   }
 
--- a/packages/protocol-kit/tests/unit/tx-options.test.ts
+++ b/packages/protocol-kit/tests/unit/tx-options.test.ts
@@ -1,0 +1,45 @@
+import chai from 'chai'
+import {
+  convertTransactionOptions,
+  createLegacyTxOptions,
+  createTxOptions
+} from '../../src/utils/transactions/utils'
+
+describe('createTxOptions', () => {
+  it('keeps nonce 0', () => {
+    chai.expect(createTxOptions({ nonce: 0 }).nonce).to.equal(0)
+  })
+
+  it('keeps a non-zero nonce', () => {
+    chai.expect(createTxOptions({ nonce: 123 }).nonce).to.equal(123)
+  })
+
+  it('omits nonce when it is not provided', () => {
+    chai.expect(createTxOptions({}).nonce).to.equal(undefined)
+    chai.expect(createTxOptions().nonce).to.equal(undefined)
+  })
+})
+
+describe('createLegacyTxOptions', () => {
+  it('keeps nonce 0', () => {
+    chai.expect(createLegacyTxOptions({ nonce: 0, gasPrice: 1 }).nonce).to.equal(0)
+  })
+
+  it('keeps a non-zero nonce', () => {
+    chai.expect(createLegacyTxOptions({ nonce: 123, gasPrice: 1 }).nonce).to.equal(123)
+  })
+
+  it('omits nonce when it is not provided', () => {
+    chai.expect(createLegacyTxOptions({ gasPrice: 1 }).nonce).to.equal(undefined)
+  })
+})
+
+describe('convertTransactionOptions', () => {
+  it('keeps nonce 0 on the EIP-1559 path', () => {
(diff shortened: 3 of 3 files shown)


devtechedge:

devtechedge:
recheckcla

devtechedge:
Gentle ping on this one.

It has been open for about ten days with CI green, and it fixes a small correctness bug: a caller-supplied `nonce` of `0` is treated as unset when execution options are converted, so the value gets dropped.

The changeset is in and the CLA check has been re-run since the last push.

Happy to retarget to `main` or reshape the fix if that makes it easier to land.

PR #1431, opened 2026-09-09
fix(protocol-kit): truncate generateHash by bytes and bump identifier to v01

### LINKED ISSUE #1424: generateHash truncates by hex character, not byte — on-chain identifier fields carry half the documented entropy
`generateHash` in `on-chain-tracking/generateOnChainIdentifier.ts` documents `size` as bytes but actually takes hex characters, giving on-chain identifier fields half the claimed entropy. Filing this as an issue rather than a PR since a real fix changes the bit pattern of every future identifier relative to already-published ones — this needs a maintainer decision on whether/how to ship it, not a silent behavior change.

Companion PR (docs + test-integrity only, no behavior change): #1423

## The bug

```ts
export function generateHash(input: string, size: number): string {
  const fullHash = keccak256(toHex(input))
  return toHex(fullHash.slice(-size)).replace('0x', '') // Take the last X bytes
}
```

`keccak256()` returns a hex string. `.slice(-size)` takes `size` **characters**, not bytes. The subsequent `toHex()` re-encodes those ASCII hex-digit characters as bytes rather than reinterpreting the underlying hash bytes — so the output round-trips back to readable ASCII text:

```
fullHash        = 0x5cf9c0e95198ffeb9ef583d208e91798eb10bbf881ec2526939f24f8049e7556
slice(-20)      = "2526939f24f8049e7556"
generateHash    = 3235323639333966323466383034396537353536
as ASCII        = "2526939f24f8049e7556"   <- identical to the sliced characters
```

## Entropy impact

For `size=3` (`platformHash`, `toolHash`, `toolVersionHash`): `16^3 = 2^12` possible values, not `256^3 = 2^24` as documented — a 4096-value space instead of ~16.7M.

For `size=20` (`projectHash`): `16^20 = 2^80`, not `2^160`.

## Why this matters in practice

`toolVersionHash` (3 bytes, 12 bits of real entropy) is derived from the tool's version string, which increments with every release. Birthday-paradox collision probability across N distinct version strings:

```
P(≥1 collision) by the 80th version, 12-bit space (actual):    0.5422
P(≥1 collision) by the 80th version, 24-bit space (documented): 0.000191
```

This isn't a fund-safety, signature, or hash-verification path — it's purely the Safe ecosystem-partner attribution/analytics identifier appended to transaction data. No revert, no fail-open/fail-closed security implication. The consequence of a collision is two different tool versions (or platforms, or projects) becoming indistinguishable in on-chain analytics.

## Why I'm not just fixing it in a PR

A byte-accurate fix would change `generateHash`'s output for every future call with the same logical input, meaning identifiers written from this point forward would carry a different bit pattern than identifiers already published on-chain for the same project/platform/tool/toolVersion combination — silently breaking any existing off-chain deduplication or attribution logic that assumes stability.

The identifier format already has a version byte for exactly this kind of change:

```ts
const identifierVersion = '00' // first version
```

A real fix should bump this to `'01'` and gate the corrected (byte-accurate) `generateHash` logic behind it, keeping `'00'`-tagged identifiers decoding with the current (nibble-based) logic for backward compatibility with anything that reads historical identifiers off-chain.

## Suggested shape (for discussion, not prescriptive)

- New `identifierVersion = '01'`.
- A corrected hash-truncation path that takes real bytes (e.g. slicing the raw digest bytes before hex-encoding, rather than slicing the hex string), used only for version `'01'` identifiers.
- Existing `generateHash`/version `'00'` path left untouched, so nothing already on-chain is affected and any downstream analytics keyed on the current format keeps working.

Happy to implement this if a maintainer confirms the version-gated shape above is the right direction.

devtechedge:
## Summary

Fixes #1424.

`generateHash` documents `size` as a byte count, but it was slicing the keccak hex **string** by character count and then running `toHex()` on those ASCII characters. That re-encodes hex digits as bytes, so each field only ever holds ASCII `'0'`-`'9'` / `'a'`-`'f'` values - about half the claimed entropy (for example `2^12` instead of `2^24` for the 3-byte fields).

## What changed

1. **Byte-accurate truncation** - take the last `size * 2` hex characters from the `0x`-prefixed digest (real bytes), and stop re-encoding with `toHex()`:
   ```ts
   return fullHash.slice(-size * 2)
   ```
2. **Bump `identifierVersion` from `00` to `01`** - so new identifiers are explicitly versioned and stay distinguishable from identifiers already published on-chain with the old nibble-based truncation. Off-chain analytics that key on the current format can keep treating `00` as before.
3. **Tests** - update e2e expectations to version `01` and assert against hard-coded byte-accurate hash vectors (not a fresh call to `generateHash` alone), so regressions in truncation are actually caught.

## Why the version bump

A silent fix to `generateHash` would change the bit pattern of every future identifier for the same project/platform/tool/toolVersion inputs relative to what is already on-chain. The format already has an `identifierVersion` byte for this kind of change, so bumping it to `01` is the safe way to ship the corrected truncation.

This path is analytics / attribution metadata appended to transaction data only - not a signature, fund-safety, or fail-open verification path. Collisions under the old entropy mainly make distinct tool versions (or platforms/projects) harder to tell apart in on-chain analytics.

## Example (size = 20)

For digest
`0x5cf9c0e95198ffeb9ef583d208e91798eb10bbf881ec2526939f24f8049e7556`:

| Step | Old (buggy) | New (byte-accurate) |
|------|-------------|---------------------|
| Slice | last 20 **chars**: `2526939f24f8049e7556` | last 20 **bytes**: `08e91798eb10bbf881ec2526939f24f8049e7556` |
| Output | `32353236...` (ASCII of those chars) | the 40 hex chars above |

## Test plan

- [ ] `yarn testing-kit test tests/e2e/onChainIdentifier.test.ts` (or equivalent protocol-kit e2e for on-chain identifier)
- [ ] Confirm identifiers now start with `5afe01`
- [ ] Confirm `generateHash(input, n)` length is always `2 * n` and is not ASCII-decodable back to a short hex substring of the digest
- [ ] Maintainers: confirm version-gated ship (`01` for new writes) matches the intended rollout for analytics consumers

Happy to adjust the versioning/compatibility shape if maintainers prefer a different approach (for example keeping a legacy helper for reproducing `00` identifiers in tests only).

### DIFF
--- a/packages/protocol-kit/src/utils/on-chain-tracking/generateOnChainIdentifier.ts
+++ b/packages/protocol-kit/src/utils/on-chain-tracking/generateOnChainIdentifier.ts
@@ -9,7 +9,8 @@ import { keccak256, toHex } from 'viem'
  */
 export function generateHash(input: string, size: number): string {
   const fullHash = keccak256(toHex(input))
-  return toHex(fullHash.slice(-size)).replace('0x', '') // Take the last X bytes
+  // Take the last `size` bytes (2 hex characters per byte). fullHash is 0x-prefixed.
+  return fullHash.slice(-size * 2)
 }
 
 export type OnChainIdentifierParamsType = {
@@ -45,7 +46,8 @@ function generateOnChainIdentifier({
   toolVersion
 }: OnChainIdentifierParamsType): string {
   const identifierPrefix = '5afe'
-  const identifierVersion = '00' // first version
+  // Version 01: byte-accurate hash truncation (see #1424). Version 00 used nibble/hex-char truncation.
+  const identifierVersion = '01'
   const projectHash = generateHash(project, 20) // Take the last 20 bytes
   const platformHash = generateHash(platform, 3) // Take the last 3 bytes
   const toolHash = generateHash(tool, 3) // Take the last 3 bytes
--- a/packages/protocol-kit/tests/e2e/onChainIdentifier.test.ts
+++ b/packages/protocol-kit/tests/e2e/onChainIdentifier.test.ts
@@ -30,14 +30,26 @@ describe('On-chain analytics', () => {
       const onChainIdentifier = generateOnChainIdentifier({ project, platform, tool, toolVersion })
 
       const identifierPrefix = '5afe'
-      const identifierVersion = '00'
+      const identifierVersion = '01'
+
+      // Hard-coded expected hashes (byte-accurate truncation, identifier version 01)
+      const expectedProjectHash = 'b239ef6cc45f806ddf9a8ae456f26a86add0878d'
+      const expectedPlatformHash = '193dea'
+      const expectedToolHash = 'bfe928'
+      const expectedToolVersionHash = 'b2972c'
 
       chai.expect(onChainIdentifier.startsWith(identifierPrefix)).to.be.true
       chai.expect(onChainIdentifier.substring(4, 6)).to.equals(identifierVersion)
-      chai.expect(onChainIdentifier.substring(6, 46)).to.equals(generateHash(project, 20))
-      chai.expect(onChainIdentifier.substring(46, 52)).to.equals(generateHash(platform, 3))
-      chai.expect(onChainIdentifier.substring(52, 58)).to.equals(generateHash(tool, 3))
-      chai.expect(onChainIdentifier.substring(58, 64)).to.equals(generateHash(toolVersion, 3))
+      chai.expect(onChainIdentifier.substring(6, 46)).to.equals(expectedProjectHash)
+      chai.expect(onChainIdentifier.substring(46, 52)).to.equals(expectedPlatformHash)
+      chai.expect(onChainIdentifier.substring(52, 58)).to.equals(expectedToolHash)
+      chai.expect(onChainIdentifier.substring(58, 64)).to.equals(expectedToolVersionHash)
+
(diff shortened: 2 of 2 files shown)


devtechedge:
Small update here: I added the changeset CONTRIBUTING.md asks for, since bumping the identifier to `01` changes the string `protocol-kit` emits.

The fix truncates `generateHash` by bytes rather than hex characters as described in issue #1424, and the e2e vectors now assert against fixed values instead of round-tripping through the same helper.

CI is green and the CLA check passes.

Happy to keep the identifier at `00` and land only the truncation if you prefer.

devtechedge:
Hello - gentle follow-up on this protocol-kit fix when you have a moment.

`generateHash` documented `size` as a byte count but was slicing the keccak hex string by character count and then re-encoding with `toHex()`, which cut the claimed entropy roughly in half.

The PR truncates by real bytes (`fullHash.slice(-size * 2)`), bumps `identifierVersion` from `00` to `01` so new identifiers stay distinguishable on-chain, and updates the e2e expectations plus the required changeset (`Fixes #1424`).

CI was green after the changeset landed, and I am glad to retarget, expand tests, or clarify the versioning note if maintainers want a different shape.

Whenever it fits the queue is fine - the branch is still ready from my end.

PR #1432, opened 2026-09-09
test(relay-kit): use disposable private keys in 4337 tests

### LINKED ISSUE #938: Use a disposable private key in the 4337 tests
## Context / issue

Currently, private keys for 4337 tests are used as CI secrets, most likely, to prevent random people to alter the state of chain in a way that break our tests hindering our operations. That introduces two issues: 
- We either have toi share a (couple) of private key secrets internally everytime some change to the implemtantion breaks the tests or we recreate the same conditions for the associated eoas/multisig which takes some effort and suffering.
- externals will never, in their lives, be able to execute (and get runtime information from) those tests.

## Proposed solution

 - Add fixtures for private keys
 - Mock the transport to avoid any consequences that could affect the result of those tests

## Expected outcome

- No data should be read from `process.env` in the tests of the 4337 module.
- No secrets data should be contained on `.github/workflows/sdk-test.yml`
- The tests for the 4337 module should pass locally and on the CI.

devtechedge:
## Summary

This updates the Safe 4337 tests so they no longer depend on shared CI private key secrets.

Previously, `Safe4337Pack` tests read `PRIVATE_KEY` and `PASSKEY_PRIVATE_KEY` from `process.env` (wired through `.github/workflows/sdk-test.yml`). That made local runs awkward for contributors and tied CI to shared secrets that had to be rotated or recreated when test state drifted.

## Changes

- Generate a disposable EOA private key in `relay-kit` test fixtures (`viem` `generatePrivateKey`) and use it as the default signer in `createSafe4337Pack`.
- Use `new WebAuthnCredentials()` without a fixed key for passkey tests, matching the protocol-kit test-utils pattern (random key when none is provided).
- Stop loading dotenv / `process.env` in the 4337 test file.
- Remove `PRIVATE_KEY` / `PASSKEY_PRIVATE_KEY` from the SDK test workflow and document that those env vars are not required in `.env.example`.
- Soften signature assertions that previously expected exact hex for a fixed owner key, and mock `isOwner` where signing with a disposable key would otherwise fail against fixture Safe addresses.

Bundler transport was already mocked in these tests, so the disposable keys stay local and do not need funded shared EOAs on Sepolia.

## Test plan

- [x] `pnpm test` in `packages/relay-kit` (includes `Safe4337Pack.test.ts`) - all passed locally
- [ ] Confirm CI `sdk-test` job passes without `TESTING_PRIVATE_KEY` / `TESTING_PASSKEY_PRIVATE_KEY`

Fixes #938

### DIFF
--- a/.github/workflows/sdk-test.yml
+++ b/.github/workflows/sdk-test.yml
@@ -38,9 +38,6 @@ jobs:
         run: pnpm build
 
       - name: Test
-        env:
-          PRIVATE_KEY: ${{ secrets.TESTING_PRIVATE_KEY }}
-          PASSKEY_PRIVATE_KEY: ${{ secrets.TESTING_PASSKEY_PRIVATE_KEY }}
         run: pnpm test
 
       - name: Upload coverage report
--- a/packages/relay-kit/.env.example
+++ b/packages/relay-kit/.env.example
@@ -1,2 +1,2 @@
-PRIVATE_KEY=
-PASSKEY_PRIVATE_KEY=
+# 4337 unit tests generate disposable private keys in fixtures.
+# No PRIVATE_KEY / PASSKEY_PRIVATE_KEY env vars are required.
--- a/packages/relay-kit/package.json
+++ b/packages/relay-kit/package.json
@@ -66,7 +66,6 @@
     "@types/jest": "^29.5.14",
     "@types/semver": "^7.7.1",
     "abitype": "^1.2.3",
-    "dotenv": "^16.4.7",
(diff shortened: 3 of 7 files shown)


devtechedge:
Wanted to flag this one while it is still small.

The 4337 tests currently hardcode private keys, and this swaps them for disposable ones generated per run, so a key committed to the suite cannot be picked up and reused.

Most of the diff is CI config and the lockfile rather than test logic.

Happy to narrow it to just the fixtures if that lowers the review cost.

